PR #20, opened 2021-11-06
Add CollectionOf Attribute for DataCollection

rwdevguy:
Currently, the `DataCollection` comment seems to require the FQN regardless of a `use` statement.

This PR provided an alternative way to declare the type of data the `DataCollection` represents:
```
class AlbumData extends Data
{
    public function __construct(
        public string $title,
        #[CollectionOf(SongData::class)]
        public DataCollection $songs,
    ) {
    }
}
```

This should not introduce a breaking change.

### DIFF
--- a/docs/advanced-usage/typescript.md
+++ b/docs/advanced-usage/typescript.md
@@ -19,7 +19,7 @@ class DataObject extends Data{
         public array $array,
         public Lazy|string $lazy,
         public SimpleData $simpleData,
-        /** @var \Spatie\LaravelData\Tests\Fakes\SimpleData[] */
+        #[CollectionOf(SimpleData::class)]
         public DataCollection $dataCollection,
     )
     {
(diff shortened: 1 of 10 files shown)


rubenvanassche:
Hi @rwdevguy,

Thanks for the PR, I'm not going to merge it since we needed a more baked in support into the package to better support data collection annotations. You can find the new docs here: https://github.com/spatie/laravel-data/blob/1.1.0/docs/as-a-data-transfer-object/collections.md#typing-data-within-your-collections

I did love the `DataCollectionOf` attribute so I've also added this to the package:

https://github.com/spatie/laravel-data/blob/1.1.0/src/Attributes/DataCollectionOf.php

Thanks for the ideas!

rwdevguy:
@rubenvanassche I completely understand! :) To be fair, I thought the naming might have been off. I really like the additional options that you've added for this.

PR #33, opened 2021-11-29
Fixed typehint in src/Casts/DateTimeInterfaceCast.php

ReDev1L:
Fixed typehint of $type variable
Added suppression of exception on DateTime creation

### DIFF
--- a/src/Casts/DateTimeInterfaceCast.php
+++ b/src/Casts/DateTimeInterfaceCast.php
@@ -5,6 +5,7 @@ namespace Spatie\LaravelData\Casts;
 use DateTimeInterface;
 use Spatie\LaravelData\Exceptions\CannotCastDate;
 use Spatie\LaravelData\Support\DataProperty;
+use Throwable;
 
 class DateTimeInterfaceCast implements Cast
 {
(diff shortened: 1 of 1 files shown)


src/Casts/DateTimeInterfaceCast.php
```diff
@@ -24,8 +24,12 @@ public function cast(DataProperty $property, mixed $value): DateTimeInterface |
             return Uncastable::create();
         }
 
-        /** @var \DateTime|\DateTimeImmutable $name */
-        $datetime = $type::createFromFormat($format, $value);
+        /** @var \DateTime|\DateTimeImmutable $type */
+        try {
+            $datetime = $type::createFromFormat($format, $value);
+        } catch (\Throwable $e) {
```
freekmurze, line 30: Import this one please 🙂

ReDev1L: Three of them? or just Throwable?
I've removed InvalidFormatException of Carbon library. Dunno how to make it cleaner.
I think thats ok, because we dont do much logic here, that can raise Exception outside of Datetime scope.

PaolaRuby: `Throwable` must be

ReDev1L: Done

freekmurze:
Good for you @rubenvanassche ?

rubenvanassche:
Looks great! Thanks!

PR #34, opened 2021-12-01
Fixes PHP-notice

ahoiroman:
Without having int as return type, we will generate notices like

```
Return type of Spatie\LaravelData\DataCollection::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /Users/ahoi/test/vendor/spatie/laravel-data/src/DataCollection.php on line 104
```


freekmurze:
Thanks!

PR #38, opened 2021-12-10
docs: change authorize to authorized

rovansteen:
👋🏻

In the docs it's mentioned that you can add a static `authorize` method to the data class, just like with a form request. However this is not correct since it checks for a method called `authorized`.  This PR fixes the documentation to be aligned with the actual code.

Just wondering if it was intentional to call it `authorized`? Since Laravel's Form Requests use `authorize`?


rubenvanassche:
Mhh, good point! The thing is, we want to reflect a Laravel form object, which uses authorize. So I'm going to change the method name to `authorize`. Thanks for noticing!

PR #52, opened 2021-12-27
PHP 8.1 for some validations

hirenkeraliya:
In PHP 8 given rule is not working but if we try in PHP 8.1 it's working.
```
#[BeforeOrEqual(Carbon::yesterday())]
```

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -4,4 +4,4 @@ weight: 3
 ---
 
 This package requires:
-- PHP 8 or higher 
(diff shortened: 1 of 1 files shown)


rubenvanassche:
I don't think we should update the docs with this, the whole package is working with PHP 8.0 perfectly fine except using some non literal values within validation rule attributes. In my opinion it's up to the PHP developer to know what can be used in a certain PHP version and what not.

PR #64, opened 2022-01-05
EnumCast and EnumTransformer

juliomotol:
This PR aims to add cast and transformer for Enums which has been introduced in PHP8.1


rubenvanassche:
Looking great! Thanks!

PR #68, opened 2022-01-10
🧪 Should be able to resolve a data with Enum properties

juliomotol:
This PR aims to add a failing test on enum casting.

In the recent changes, `Enum` has been added as a built in type:

https://github.com/spatie/laravel-data/blob/df378da8887c5290f0972f18856c87dcb861377d/src/Support/DataProperty.php#L237-L241

https://github.com/spatie/laravel-data/blob/df378da8887c5290f0972f18856c87dcb861377d/src/Resolvers/DataFromArrayResolver.php#L44-L46

These lines prevents from casting the `Enum` type even if `WithCast` is explicitly given or `data.casts.\BackedEnum` is set.

### DIFF
--- a/tests/DataTest.php
+++ b/tests/DataTest.php
@@ -721,20 +721,4 @@ class DataTest extends TestCase
         $this->assertInstanceOf(IntersectionTypeData::class, $data);
         $this->assertEquals($collection, $data->intersection);
     }
-
-    /** @test */
-    public function it_can_include_enum_data()
-    {
-        $this->onlyPHP81();
-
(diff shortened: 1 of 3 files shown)


aidan-casey:
Almost certainly the `shouldBeCasted` method should turn into:

```php
private function shouldBeCasted(DataProperty $property, mixed $value): bool
{
    return ! is_null($property->castAttribute());
}
```

The current implementation would not allow any built-in types, such as dates in string form (e.g., `2021-01-01 12:0:0`) to be casted.

juliomotol:
Closing the PR in favor of https://github.com/spatie/laravel-data/pull/75

PR #71, opened 2022-01-13
Allow custom transformers to target Data & DataCollections

LukeAbell:
I have a use case that I want to use a custom transformer targeting a data class- can we re-order this logic to run the auto data/data collection after looking for custom transformers?


rubenvanassche:
Yeah, gonna take a look at that now since it messes with Lazy properties.

rubenvanassche:
@LukeAbell, this should be fixed with 1.4.2, which I will tag later this day!

LukeAbell:
@rubenvanassche Thanks!

PR #75, opened 2022-01-18
Casting Fix: Removes simple types check when casting

aidan-casey:
This specific issue has been a big problem for us when attempting to use this package with PHP 8.1 enums. Essentially the current check for simple types avoids taking an enum value (e.g., "foo") and casting it into an enum instance because the enum value is a string and returning true on `isSimpleType()`.

I believe the best approach to fixing this is removing the check for a simple type entirely. I imagine people may run into further issues with that down the road when writing custom casters.

See also: #68.


aidan-casey:
You bet!

PR #80, opened 2022-01-24
fix for aborting value assignment after a false boolean

hxnk:
An arrow function was used when iterating over class properties; whenever a boolean type with value `false` was assigned, the loop would break and subsequent properties were not assigned.

Using an anonymous function without returning anything solves this.

I've added a test to proof the bug/fix.


rubenvanassche:
Thanks! Good catch!

PR #83, opened 2022-01-26
Auto-generate data classes for models

cerbero90:
This PR introduces an Artisan command to automatically generate data classes for models:

```
php artisan make:data App/User
```

The properties of models are scanned from the DB and their types converted into PHP types.

The relationships of models are detected from the models themselves using regular expressions. If a relationship relates to many models, an attribute `#[DataCollectionOf]` is included as well.

Properties, relationships and `use` statements are finally added to the generated data class, that will look like this:

```php
<?php

namespace App\Data;

use Carbon\Carbon;
use Spatie\LaravelData\Attributes\DataCollectionOf;
use Spatie\LaravelData\Data;
use Spatie\LaravelData\DataCollection;

/**
 * The data class of the User model.
 *
 */
class UserData extends Data
{
    /**
     * Instantiate the data class.
     *
     */
    public function __construct(
        public int $id,
        public string $username,
        public string $email,
        public string $password,
        public Carbon $createdAt,
        public Carbon $updatedAt,
        #[DataCollectionOf(PostData::class)]
        public DataCollection $posts,
        public TeamData $team,
    ) {}
}
```

The command also offers an option to generate data classes for related models recursively, so that all data classes of a model and its relationships can be generated in one go:

```
php artisan make:data App/User --recursive

php artisan make:data App/User -r
```

Data classes that already exists are not overwritten, unless we add the `--force` option:

```
php artisan make:data App/User --force

php artisan make:data App/User -f
```

We can customize how the generated data classes look like by publishing and updating the data class stub:

```
php artisan vendor:publish --tag=laravel-data-stubs
```

We can also decide the fully qualified class name (FQCN) of the generated data classes. By default the name is based on the model:
- `App\User` generates `App\Data\UserData`
- `App\Domain\Users\User` generates `App\Domain\Users\Data\UserData`.

To change this behavior, we can create a custom `DataQualifier` class and configure it in the `data.qualifier` setting.


freekmurze:
Seem very cool! Thanks for your work on this!

cerbero90:
My pleasure, thank you for this incredibly useful package 👏

rubenvanassche:
Hi @cerbero90,

Thanks for your PR, sorry for the late response. I've been thinking quite a lot about this PR last week. I think this wouldn't be a core feature for the package. Since it adds an external dependency which I want to avoid as much as possible. And it isn't really a core functionality of the data package.

Also, the `cerbero/eloquent-inspector` needs some extra functionality like attribute getters/setters and casts to properly work.

That being said, I really like the idea and I think it is a great asset for the the package itself. Maybe you could make an external package that adds this functionality. I would happily add some paragraphs to the data documentation that describes how to use such package.

One last note, if you extend `cerbero/eloquent-inspector` with support for attribute getters/setters, then please let me know. I was thinking about implementing such a package to be used with typescript-transformer so it can automatically transform models to TypeScript.

cerbero90:
No worries, @rubenvanassche I totally understand :)

It makes sense to extract this logic into a separate package, I'll let you know if/when I'll create such package.

Also getters, setters and casts are in planning for Eloquent Inspector, will keep you in the loop.

Thanks for taking the time to look into this 👍

PR #94, opened 2022-02-12
Allow using default password config

abenerd:
https://laravel.com/docs/9.x/validation#defining-default-password-rules

The Password validation rules allows one to set a default configuration by providing a callback that returns a configured password class.

This PR allows using this default validation config.


rubenvanassche:
Thanks!

PR #101, opened 2022-02-15
Solves binding issues with Laravel Octane.

aidan-casey:
The current implementation is not compatible with Laravel Octane because it incorrectly binds the request. This PR allows the package to work within Octane.


rubenvanassche:
Thanks Casey!

PR #106, opened 2022-02-18
Add support for stdClass payload casting

LaravelFreelancerNL:
In the previous version trying to create a Data object from a stdClass object would error out.

This PR adds support for stdClass objects by simply casting it to an array before proceeding with the DataFromArrayResolver.

This is particularly useful for casting nested objects which, in my use case, are returned from a document model database.


rubenvanassche:
Thanks!

PR #110, opened 2022-03-04
V2

rubenvanassche:
See #84


docs/getting-started/quickstart.md
```diff
@@ -5,7 +5,7 @@ weight: 1
 
 In this quickstart, we'll guide you through the most important functionalities of the package and how to use them. 
 
-First, you should [install the package](https://spatie.be/docs/laravel-data/v1/installation-setup).
+First, you should [install the package](https://spatie.be/docs/laravel-data/v2/installation-setup).
 
 We're going to create a blog with different posts so let's get started with the `PostData` object. A post has a title, some content, a status and a date when it was published:
```
freekmurze, line 10: Can you modernise the enum example to use native enums?

rubenvanassche: I think we want to keep it that way since it explains how Transformers and Casts work, but maybe the QuickStart need a complete rewrite with a better example 🤔

PR #112, opened 2022-03-14
`ValidationAttribute` should be extended

ace18:
`ValidationAttribute` should be extended not implemented, it's an abstract class not an interface


rubenvanassche:
Good catch, thanks!

PR #114, opened 2022-03-18
Add Dependency Injection for `rules`, `messages`, and `attributes` methods

bennett-treptow:
This PR aims to add dependency injection for the `rules()`, `messages()`, and `attributes()` methods. This will allow users to add parameters to any of these methods that will inject useful classes like the current Request object to make more fine-tuned rules, messages, and attributes. Of course, anything could be injected that does not rely on the current request as well, so that the data object can still use `SongData::validate(['title' => 'Never gonna give you up', 'artist' => 'Rick Astley']);` without throwing an error about it not being able to resolve the request from the container.

Examples:
```php
use \Illuminate\Http\Request;
class SongData extends Data
{
    public function __construct(
        public string $title,
        public string $artist,
    ) {
    }
    
    public static function rules(Request $request): array
    {
        return [
            'title' => [new RequiredIf($request->input('new_title_required')), new StringType()],
            'artist' => [new Required(), new StringType()],
        ];
    }
}
```

### DIFF
--- a/docs/as-a-data-transfer-object/request-to-data-object.md
+++ b/docs/as-a-data-transfer-object/request-to-data-object.md
@@ -126,6 +126,28 @@ class SongData extends Data
 ```
 Rules defined within the `rules` method will always overwrite automatically generated rules.
 
+You can even add dependencies to be automatically injected:
+```php
+use SongSettingsRepository;
+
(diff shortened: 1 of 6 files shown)


rubenvanassche:
Thanks!

PR #115, opened 2022-03-30
feat: `except` and `includeWhen`

innocenzi:
This PR adds support for conditionally omitting properties from a DTO. Please refer to https://github.com/spatie/laravel-data/discussions/111 for the context. It also partially addresses https://github.com/spatie/laravel-data/discussions/5, and https://github.com/spatie/laravel-data/discussions/84 (v2) since it's on the "consider" list.

An `excludeWhen` and an `except` method have been added through a trait, which makes it possible to either define how to exclude properties conditionnally, or to exclude them at runtime.

The documentation has been updated too.

## Examples

**Omitting properties conditionally:**
```php
class UserData extends Data
{
    public function __construct(
        public int $id,
        public string $name,
    ) {
    }
    
    public function excludeWhen(): array
    {
        return [
          'id' => !auth()->user()?->is_admin
        ];
    }
}
```

**Omitting properties at runtime:**
```php
UserData::from($user)->exclude('id', 'password');
```

### DIFF
--- a/docs/as-a-resource/from-data-to-resource.md
+++ b/docs/as-a-resource/from-data-to-resource.md
@@ -351,3 +351,33 @@ Now each transformed data object contains an `endpoints` key with all the endpoi
     }
 }
 ```
+
+## Omitting properties
+
+It is possible to conditionnaly omit properties, which can be useful when not wanting to expose certain properties. 
+
+To achieve this, you can extend `excludeWhen` method, which should return an array. In the following example, `id` will be excluded from the resource unless the authenticated user is an administrator.
+
+```php
+class UserData extends Data
+{
+    public function __construct(
+        public int $id,
+        public string $name,
+    ) {
+    }
+    
+    public function excludeWhen(): array
+    {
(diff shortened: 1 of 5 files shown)


rubenvanassche:
Hi @innocenzi, 

Thanks for your PR, this is indeed already implemented within v2 which will release in a few weeks. I think it is better to wait for that release so the behavior is consistent between both versions of the package.

innocenzi:
Hey, this is great news, I'm totally fine with that. Thanks! :)

innocenzi:
@rubenvanassche By the way, when you say it's implemented, I have seen `except` and `only` have been added, but is there an equivalent of that `excludeWhen` functionality?

Because having to use `except` everywhere would be inconvenient.

rubenvanassche:
Not at the moment, feel free to PR it to the v2 branch!

innocenzi:
> Not at the moment, feel free to PR it to the v2 branch!

Done! :)

PR #119, opened 2022-04-07
feat: make nullable properties optional

innocenzi:
This PR makes nullable properties optional, just like `Lazy` already does. The rationale behind this is that in TypeScript, we almost never want to have actual `null` types. We prefer `undefined` because properties can then be omitted, and most operations return `undefined`, not `null` (eg. [`find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find), [`pop`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop), etc...).

I talked about it a bit in this issue: https://github.com/spatie/typescript-transformer/issues/23

I think this should also be addressed in `typescript-transformer` but the only place I can see it being used is in `DtoTransformer`. I can PR there too if this one is accepted.

### DIFF
--- a/src/Support/TypeScriptTransformer/DataTypeScriptTransformer.php
+++ b/src/Support/TypeScriptTransformer/DataTypeScriptTransformer.php
@@ -56,7 +56,7 @@ class DataTypeScriptTransformer extends DtoTransformer
                     $property->getDeclaringClass()->getName()
                 );
 
-                return $this->isPropertyLazy($property)
+                return $this->shouldBeOptional($property)
                     ? "{$carry}{$property->getName()}?: {$transformed};" . PHP_EOL
                     : "{$carry}{$property->getName()}: {$transformed};" . PHP_EOL;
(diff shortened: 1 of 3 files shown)


rubenvanassche:
Hi @innocenzi,

Thanks for the PR wide at Spatie do use the null type within our TypeScript code so this is a change we cannot do. We do however add an `Undefined` type in the package v2 version. The plan is to make types that have this type as a union undefinable in TypeScript just like your PR.

I do not think we can merge concepts of `null` and undefinable properties since they are quite different. That's why I'm closing this PR.

innocenzi:
If I may ask, how do you use `null` in your code? Can you give me a practical example? I'm struggling to find a case where you would want that.

I'm fine with the PR being closed as long as there is a way to make them undefineable, though I think the DX of using PHP's nullable would've been better.

PR #121, opened 2022-04-07
Change `DataConfig` to scoped binding (instead of singleton)

AlexVanderbist:
Prevents accidentally contaminating subsequent requests if the `DataConfig` should get changed somewhere in the request


rubenvanassche:
Not sure if we actually need to do this? Since the `DataConfig` only contains configuration very specific to the code and not based upon any request data.

PR #123, opened 2022-04-26
feat(rules): support `$payload` as a dependency

innocenzi:
This PR allows passing `$payload` as a parameter to the static `rules` method from a `Data` object.

This is very useful when needing to conditionally apply rules based on the data you get. This is already possible by injecting `Request` in the method, but this only works when working in the context of the request. 

For instance, when using `SongData::validate(['name' => 'Wasted Youth'])`, a `Request` object will not contain the `name`.

Instead, you can now do the following:

```php
class PaymentData extends Data
{
    public function __construct(
        public string $payment_method,
        public ?string $paypal_email,
    ) {
    }
    
    public static function rules(array $payload): array
    {
        return [
            'payment_method' => ['required'],
            'paypal_email' => Rule::requiredIf($payload['payment_method'] === 'paypal'),
        ];
    }
}
```

Note that the `$payload` variable name is important, it's a named parameter. It can obviously be omitted too.

### DIFF
--- a/docs/as-a-data-transfer-object/request-to-data-object.md
+++ b/docs/as-a-data-transfer-object/request-to-data-object.md
@@ -148,6 +148,27 @@ class SongData extends Data
 }
 ```
 
+Additionally, if you need to access the data payload, you can use `$payload` parameter:
+
+```php
+class PaymentData extends Data
+{
+    public function __construct(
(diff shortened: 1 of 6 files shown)


rubenvanassche:
Thanks!

PR #126, opened 2022-04-29
feat: `includeWhen` and `excludeWhen`

innocenzi:
This PR is a follow-up of https://github.com/spatie/laravel-data/pull/115, as asked in the comments. 

Its purpose is to have an in-DTO way of conditionally excluding or including data:

```php
class UserData extends Data
{
    public function __construct(
        public int $id,
        public string $name,
    ) {
    }
    
    public function excludeWhen(): array
    {
        return [
          'id' => !auth()->user()?->is_admin
        ];
    }
}
```

Note: this PR also adds support for `includeWhen`, since the implementation was similar - but I'm unsure of the naming for this one, it is kind of confusing, any input there?

### DIFF
--- a/src/Concerns/IncludeableData.php
+++ b/src/Concerns/IncludeableData.php
@@ -53,17 +53,40 @@ trait IncludeableData
         return $this;
     }
 
+    public function includeWhen(): array
+    {
+        return [];
+    }
+
+    public function excludeWhen(): array
+    {
+        return [];
+    }
+
     public function getPartialTrees(): PartialTrees
     {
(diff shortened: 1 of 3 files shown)


innocenzi:
Hey @rubenvanassche, I was eagerly waiting for this feature because we need to not send properties to the front-end (Inertia) under certain conditions. 

It seems that you modified the feature to no longer be able to define that inside the Data class, and instead only allow that by chaining `exceptWhen` or `onlyWhen`, which is a different feature. 

Is there a reason you changed that? Can I PR that again?

PR #127, opened 2022-04-29
feat: support optional properties

innocenzi:
This PR is a second take at https://github.com/spatie/laravel-data/pull/119, that does not replace the default behavior but supports the newly configurable one. 

It depends on https://github.com/spatie/typescript-transformer/pull/30 and https://github.com/spatie/laravel-typescript-transformer/pull/15. Both of these should be merged before this PR (tests will not pass anyway).

What's important:
- The `#[Optional]` attribute from `typescript-transformer` is taken into account
- The new `transform_null_to_optional` option from `laravel-typescript-transformer` is taken into account
- The current behavior *do not changes* unless one of the above is true

Example:

```php
class UserData
{
    public function __construct(
        #[Optional]
        public int $id,
        public string $first_name,
    ) {
    }
}
```

```ts
interface UserData {
  id?: number
  first_name: string
}
```

The `Optional` attribute above would not be necessary if `config/typescript-transformer.php`'s `transform_null_to_optional` is set to `true` (which is `false` by default to preserve current behavior).

cc @rubenvanassche


innocenzi:
After reading the new code (that's a big refactor!), partially. The most important feature of that PR won't work:
- The new `transform_null_to_optional` option from `laravel-typescript-transformer` is taken into account

Though I'm not sure how to get the config in `DataType` yet.

I also wanted to support `#[Optional]` for consistency with `spatie/laravel-typescript-transformer`.

---

Additional note, how do you feel about renaming `Undefinable` to `Optional`? That's more in line with what it is and what it does, in my opinion. It's also the term used in TypeScript.

innocenzi:
@rubenvanassche should I open again targeting `main`?

rubenvanassche:
Yes! Create plan! typescript-transformer and a brand new typescript-transformer v3 is the next big thing I want to work on!

PR #130, opened 2022-05-07
feat: support multiple date formats

innocenzi:
This PR is the same as https://github.com/spatie/laravel-data/pull/120, except it targets the `v2` branch and makes a breaking change by updating the configuration format. 

I realized that when serializing a date, having multiple formats didn't make sense. So I updated the configuration to have a `from` and `to` option, where `from` is an array of supported input date formats, and `to` the format to which a date will be serialized into.

---

Initial PR text:

> This PR adds support for multiple date formats, instead of just one. When working with multiple applications (web front-end, APIs, etc), it's easier to accept multiple instead of requiring a specific one.
>
> This change is fully backwards-compatible, but allows the `date_format` configuration option to be an array. 
>
> The `DateTimeInterfaceCast` will now loop through the configured formats and use the first one that doesn't fail. If all of them fail, it will throw as usual.
>
> Additionnally, I added the most-commonly used formats as the defaults: `ATOM`, `ISO8601`, and `Y-m-d H:i:s` (the same but without the period).
>
> Partially related: https://github.com/php/php-src/pull/8322

### DIFF
--- a/config/data.php
+++ b/config/data.php
@@ -2,7 +2,9 @@
 
 return [
     /*
-     * The package will use this date format when working with dates through the app
+     * The package will use this format when working with dates. If this option
+     * is an array, it will try to convert from the first format that works,
+     * and will serialize dates using the first format from the array.
      */
     'date_format' => DATE_ATOM,
 
--- a/docs/as-a-data-transfer-object/casts.md
+++ b/docs/as-a-data-transfer-object/casts.md
@@ -146,4 +146,3 @@ class SongData extends Data
 ## Creating your own casts
 
 It is possible to create your casts. You can read more about this in the [advanced chapter](/docs/laravel-data/v1/advanced-usage/creating-a-cast).
-
--- a/docs/installation-setup.md
+++ b/docs/installation-setup.md
@@ -20,7 +20,9 @@ This is the contents of the published config file:
 ```php
 return [
     /*
-     * The date format to be used when converting a DateTimeInterface from and to json
+     * The package will use this format when working with dates. If this option
+     * is an array, it will try to convert from the first format that works,
+     * and will serialize dates using the first format from the array.
      */
     'date_format' => DATE_ATOM,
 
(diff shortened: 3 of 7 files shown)


rubenvanassche:
Hi @innocenzi, looks like your branch is a bit out of sync. I'm not entirely sure if we want to implement such a feature since you probably decide to use one date time format through your complete application. If you need to divert from that format then you probably want to manually correct it using the `format` property in the attributes.

I don't think that looping through a list of possible formats hoping to find a correct one is a solution that is really deterministic.

But correct me if I'm wrong 😄

innocenzi:
Hey, this PR indeed is outdated since your latest refactor. Love your code by the way. 

Related to your thoughts about the PR, I explained it in #120, but let me TL;DR: basically, when your DTO is used through an API, a web app, and other clients, it's very useful to accept multiple date formats.

For instance, by default, JavaScript's `Date`'s UTC method will use the `Y-m-d\TH:i:s.u\Z` format. MySQL will use `Y-m-d H:i:s` as their format. Our iOS developer uses `Y-m-d\TH:i:sO` in our two apps.

Depending on the interface/client, multiple formats can be used, and it's useful to easily accept all of them instead of forcing them to change everything to use one specific format. 

That's the use case this PR is trying to solve. Maybe the implementation isn't exactly right, but I do think the use case is legitimate. What do you think?

rubenvanassche:
Hmm, I get your point but I think it's a bit too heavy for users who don't need this. Could we change the config to be one date by default without the `from` and `to` options? You would then be able to define an array which will be looped upon when creating a date. When transforming a date we would just use the first format defined if an array is given

innocenzi:
Okay, that's basically reverting my latest changes, haha. That was my initial approach in https://github.com/spatie/laravel-data/pull/120. 

I updated the PR, and also kept the default configuration as it is now. 👍

rubenvanassche:
Thanks!

PR #147, opened 2022-07-01
Use protected properties everywhere 

ralphjsmit:
This PR changes the last `private` properties to be `protected` properties. This is incredibly useful when people need to override a class or sth. In addition, there were already quite some places where the `protected` visibility was already used, so this makes it consistent. 

My use case was that I wanted to override the `transform()` method on the `DataCollection` class, because I needed to use an overrided `DataCollectionTransformer`. The regular `DataCollectionTransformer` uses a custom array structure for paginated JSON responses, which is not the same as the default Laravel paginated array structure. We don't want to release breaking changes to our API-users, so I had to override `DataCollectionTransformer` class.

However, when all the properties are `private`, overriding is almost impossible. 

It would be greatly appreciated if this could be merged faster than later. Thanks a lot!

### DIFF
--- a/src/DataCollection.php
+++ b/src/DataCollection.php
@@ -30,14 +30,14 @@ class DataCollection implements Responsable, Arrayable, Jsonable, JsonSerializab
     use ResponsableData;
     use IncludeableData;
 
-    private ?Closure $through = null;
+    protected ?Closure $through = null;
 
-    private ?Closure $filter = null;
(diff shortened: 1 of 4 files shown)


rubenvanassche:
We're currently moving to Laravel-data v2, I hope to release it today! I've just updated the whole codebase of the v2 branch to only use protected methods and properties. I'm merging this one so v1 also has these protected properties and methods.

PR #154, opened 2022-07-08
fix(ts): use mapped property names in generated types

innocenzi:
This PR fixes the generated TypeScript types not properly using the defined mapped property names.

For instance:

```php
#[MapOutputName(SnakeCaseMapper::class)]
class ContractData extends Data
{
    public function __construct(
        public string $name,
        public string $recordCompany,
    ) {
    }
}
```

Would output:

```ts
interface ContractData {
  name: string
  recordCompany: string
}
```

After this PR, the correct interface is generated:

```ts
interface ContractData {
  name: string
  record_company: string
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
     },
     "scripts": {
         "analyse": "vendor/bin/phpstan analyse",
-        "test": "./vendor/bin/paratest --no-coverage",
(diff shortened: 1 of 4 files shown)


rubenvanassche:
Thanks!

PR #157, opened 2022-07-08
feat: add `make:data` artisan command

innocenzi:
I often reach for it but it doesn't exist, so this PR adds a `make:data` artisan command. Nothing special there, a simple generator command with a stub.


rubenvanassche:
Thanks!

PR #162, opened 2022-07-11
Allow arrays in query parameters for partials

erikgaal:
When building the partials tree from the query parameters the only supported syntax is comma separated: `?include=foo,bar`

In a project I'm working on we're already using that key, but with array syntax: `?include[]=foo&include[]=bar`

This PR add support for both syntaxes.


erikgaal:
All done! @rubenvanassche

rubenvanassche:
Thanks, looking awesome like always!

PR #164, opened 2022-07-13
Feature/extended partials

rubenvanassche:
Adds support for extended partials which can be defined on a class. Rework of #155.

```php
class AlbumData extends Data
{
    public function __construct(
        public string $title,
        #[DataCollectionOf(SongData::class)]
        public Lazy|DataCollection $songs,
    ) {
    }
    
    public function includes(): array
    {
        return [
            'songs' => count($this->songs) > 0,
        ];
    }
}
```


innocenzi:
Hey @rubenvanassche - I tried this PR for my use case, here is my feedback:

`includes` and `excludes` don't seem to work, or at least I don't understand what they're supposed to do. In the example below, `id` is still included:

```php
final class EmployeeData extends Data
{
    public function __construct(
        public readonly int $id,
        public readonly string $full_name,
        public readonly string $first_name,
        public readonly ?string $last_name,
        public readonly string $trigram,
    ) {
    }

    public function excludes(): array
    {
        return [
            'id' => true,
        ];
    }
}
```

I tried `except` (as a property) and `excepts` (as a function), and they do work. I'm a bit skeptical about the method name `onlys`, but I'm fine as long as I can conditionally exclude whatever from inside the data object :p

rubenvanassche:
That's true: `includes` and `excludes` refer to `include` and `exclude` of lazy properties. So they won't work like `only` and `except`. See the docs for more information about this.

As for `onlys` also not a fan of the name but I don't have any better idea at the moment.

rubenvanassche:
Did some additional renaming for consistency.

PR #174, opened 2022-07-20
[Docs] change create to from

Nir-An:
create is not an option, so maybe it should be from


rubenvanassche:
Thanks, that's something left from the earlier versions of the package 😄

Nir-An:
> Thanks, that's something left from the earlier versions of the package smile

I like this package. 

Thanks to you.

PR #184, opened 2022-08-03
Conditional return types for collections

rovansteen:
Resolves #158 

This adds conditional return types for collections so that `phpstan` knows which collection type is being returned (`DataCollection`, `PaginatedDataCollection`, `CursorPaginatedDataCollection`)

There is [an issue](https://github.com/phpstan/phpstan/issues/7740) and that is that it doesn't seem to properly work with interfaces, so (for now) it checks for the `CursorPaginator` and `Paginator` classes rather than interfaces.
Once that bug has been fixed we can update the conditional return type to check for the contracts, but it still might be worth getting this in before that.

### DIFF
--- a/src/Contracts/BaseData.php
+++ b/src/Contracts/BaseData.php
@@ -30,7 +30,7 @@ interface BaseData
     /**
      * @param \Illuminate\Support\Enumerable|array|\Illuminate\Pagination\AbstractPaginator|\Illuminate\Contracts\Pagination\Paginator|\Illuminate\Pagination\AbstractCursorPaginator|\Illuminate\Contracts\Pagination\CursorPaginator|\Spatie\LaravelData\DataCollection $items
      *
-     * @return \Spatie\LaravelData\DataCollection|\Spatie\LaravelData\CursorPaginatedDataCollection|\Spatie\LaravelData\PaginatedDataCollection
+     * @return ($items is \Illuminate\Pagination\AbstractCursorPaginator|\Illuminate\Pagination\CursorPaginator ? \Spatie\LaravelData\CursorPaginatedDataCollection : ($items is \Illuminate\Pagination\Paginator|\Illuminate\Pagination\AbstractPaginator ? \Spatie\LaravelData\PaginatedDataCollection : \Spatie\LaravelData\DataCollection))
      */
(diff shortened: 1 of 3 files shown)


erikgaal:
I think it's fine to support _just_ Laravel's native paginators for now as that will resolve this issue for the majority of the users.

rubenvanassche:
Is the `types/collection.php` file required?

rovansteen:
@rubenvanassche it's not required for the functionality. It's kind of like a test file for types that gets checked by phpstan to ensure that it works and keeps working in the future. It's similar to how Laravel does this: https://github.com/laravel/framework/tree/9.x/types

rubenvanassche:
Didn't know about that one, great! Thanks!

src/Contracts/BaseData.php
```diff
@@ -30,7 +30,7 @@ public static function from(mixed ...$payloads): object;
     /**
      * @param \Illuminate\Support\Enumerable|array|\Illuminate\Pagination\AbstractPaginator|\Illuminate\Contracts\Pagination\Paginator|\Illuminate\Pagination\AbstractCursorPaginator|\Illuminate\Contracts\Pagination\CursorPaginator|\Spatie\LaravelData\DataCollection $items
      *
-     * @return \Spatie\LaravelData\DataCollection|\Spatie\LaravelData\CursorPaginatedDataCollection|\Spatie\LaravelData\PaginatedDataCollection
+     * @return ($items is \Illuminate\Pagination\AbstractCursorPaginator|\Illuminate\Pagination\CursorPaginator ? \Spatie\LaravelData\CursorPaginatedDataCollection : ($items is \Illuminate\Pagination\Paginator|\Illuminate\Pagination\AbstractPaginator ? \Spatie\LaravelData\PaginatedDataCollection : \Spatie\LaravelData\DataCollection))
```
IAmRGroot, line 33: Would it be possible to add the TKey and TValue in the return types? 

Something like `\Spatie\LaravelData\DataCollection<int,static>` ?

This prevents me from having to do it myself in my Data objects

rubenvanassche: How would this look like? I'm all in for better types so if you can send in a Pr I'll probably merge it! 😄

PR #193, opened 2022-08-16
feat: allow direct pipeline call

innocenzi:
This PR adds a way to transform something into a DTO without using the custom transform methods. 

I often need to transform models but modify some of their properties during the process. For instance, a model using HashIDs should not have its `id` used in the DTO, but its HashID. 

Currently, I use a custom `from` method, and I call `toArray()` on the model:

```php
public static function fromDummyModel(DummyModel $model): static
{
    return static::from([
        ...$model->toArray(),
        'id' => $model->getRouteKey()
    ]);
}
```

The issue is that this will not take nested DTOs into account. If my model has a relation, and that relation is setup as a DTO, the custom creation methods for these nested DTOs are never called, because the relations are serialized in the first `toArray` call. 

To fix that, this PR adds a `throughPipeline` method that will pass the given value through the transform pipeline while ignoring custom transform methods. This method can be called inside a custom transform method, fixing the issues above.

```php
public static function fromDummyModel(DummyModel $model)
{
    return static::from([
        ...self::throughPipeline($model)->toArray(),
        'id' => $model->getRouteKey(),
    ]);
}
```

If you have a better name than `throughPipeline`, feel free to change it. I couldn't come up with a better idea. 😅

### DIFF
--- a/docs/advanced-usage/pipeline.md
+++ b/docs/advanced-usage/pipeline.md
@@ -60,4 +60,4 @@ The `handle` method has several arguments:
 
 When using a magic creation methods, the pipeline is not being used (since you manually overwrite how a data object is
 constructed). Only when you pass in a request object a minimal version of the pipeline is used to authorize and validate
-the request.
+the request. If you need to ignore the custom creation methods, you can call `throughPipeline` on the data object.
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -61,6 +61,11 @@ parameters:
 			path: src/Data.php
 
 		-
+			message: "#^Method Spatie\\\\LaravelData\\\\Data\\:\\:throughPipeline\\(\\) should return static\\(Spatie\\\\LaravelData\\\\Data\\) but returns Spatie\\\\LaravelData\\\\Contracts\\\\BaseData\\.$#"
+			count: 1
+			path: src/Data.php
+
+		-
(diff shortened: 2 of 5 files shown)


innocenzi:
I'm thinking of renaming `throughPipeline` to `make`, what do you think? Too vague?

rubenvanassche:
Couldn't you use a cast in this case? That would use the hashId instead of the id.

innocenzi:
Hey, I think I could in that specific use case but I also think this is useful. 

I had multiple situations where I needed to convert half the DTO in a normal way, and perform diverse operations in the custom creation method - in that case, I'm not even sure adding casts would work.

innocenzi:
Here is an example where the first line would benefit from `throughPipeline` and the rest would be cumbersome to describe elsewhere.

```php
public static function fromModel(BookingEstimate $estimate): self
{
    return self::from([
        ...$estimate->getAttributes(),
        'operator' => Lazy::create(fn () => OperatorData::from($estimate->operator)),
        'booking' => Lazy::create(fn () => BookingData::from($estimate->booking)),
        'aircraft' => Lazy::create(fn () => AircraftData::from($estimate->aircraft)),
        'operator' => Lazy::create(fn () => OperatorData::from($estimate->operator)),
        'legs' => Lazy::create(fn () => LegData::collection($estimate->legs)),
        'total_flight_time' => $estimate->legs_sum_flight_time ?? $estimate->totalFlightTime(),
        'is_urgent' => $estimate->isUrgent(),
    ]);
}
```

rubenvanassche:
Hmm, I get your point. But I'm gonna change the implementation a bit

rubenvanassche:
Thanks

innocenzi:
> Hmm, I get your point. But I'm gonna change the implementation a bit

As long as the functionality remains, I don't mind at all. Thanks for your time ❤️

PR #196, opened 2022-08-18
refactor(var-dumper): use `AbstractCloner` instead of reflection

innocenzi:
The current implementation does not work sometimes when the handler isn't the one you expect and does not have the `cloner`:

image

image

My whole test suit fails right now. This PR updates the implementation to simply use the `AbstractCloner`.


rubenvanassche:
Thanks

PR #197, opened 2022-08-18
Fix: Validators can be applied for nested data

sash:
Attribute validators cannot be applies to Data and Collections - there are defaults that are used instead. This fix will get the validators from the basic property and merge any tested validators with them


rubenvanassche:
Could you please provide a test which tests this behavior that way we make sure the code keeps working in the future!

rubenvanassche:
This is also targeted at laravel-data v1 and I think fixed in v2.

PR #199, opened 2022-08-21
feat: return property with or without mapping the name

Nir-An:
Add an ability to return property with or without mapping the name:

```php
class ContractData extends Data
{
    public function __construct(
        public string $name,
        #[MapOutputName('record_company')]
        public string $recordCompany,
    ) {
    }
}
$songData = ContractData::from(['name' => 'Rick' , 'recordCompany' => 'RCA'])

$songData->all();
//  result: ['name' => 'Rick', record_company => 'RCA' ], // keep mapping by default 

$songData->toArray();
//  result: ['name' => 'Rick', record_company => 'RCA' ], // keep mapping by default 


$songData->transform(transformValues: true, mapPropertyNames:true //default value );
//  result: ['name' => 'Rick', record_company => 'RCA' ], 


$songData->transform(transformValues: true, mapPropertyNames:false );
//  result: ['name' => 'Rick', recordCompany => 'RCA' ],

```
~~With this PR the `all()` function   can return properties with the original names~~

Update Docs
Add tests

related to discussion  #183


src/Contracts/TransformableData.php
```diff
@@ -20,6 +20,7 @@ public function jsonSerialize(): array;
 
     public function transform(
         bool $transformValues = true,
+        bool $mapPropertyNames = true,
```
rubenvanassche, line 23: Can you add this behind $wrapExecutionType? That way we do not have a breaking change.

src/Concerns/TransformableData.php
```diff
@@ -8,7 +8,7 @@ trait TransformableData
 {
     public function all(): array
     {
-        return $this->transform(transformValues: false);
+        return $this->transform(transformValues: false, mapPropertyNames: false);
```
rubenvanassche, line 11: Let's keep the functionality of all as is for now since it would be a breaking change!

rubenvanassche (changes requested):
Great PR! Some small remarks

src/Concerns/BaseDataCollectable.php
```diff
@@ -34,11 +35,13 @@ public function count(): int
      */
     public function transform(
         bool $transformValues = true,
+        bool $mapPropertyNames = true,
```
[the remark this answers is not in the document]
Nir-An: @rubenvanassche 
if `behind` means inserting the logic inside wrapExecutionType is kind of possible but very not recommended

src/Transformers/DataTransformer.php
```diff
@@ -217,18 +218,12 @@ protected function resolvePropertyValue(
...
-            $value instanceof BaseData && $this->wrapExecutionType === WrapExecutionType::Enabled => WrapExecutionType::TemporarilyDisabled,
-            $value instanceof BaseData && $this->wrapExecutionType === WrapExecutionType::Disabled => WrapExecutionType::Disabled,
-            $value instanceof BaseData && $this->wrapExecutionType === WrapExecutionType::TemporarilyDisabled => WrapExecutionType::TemporarilyDisabled,
-            $value instanceof BaseDataCollectable && $this->wrapExecutionType === WrapExecutionType::Enabled => WrapExecutionType::Enabled,
-            $value instanceof BaseDataCollectable && $this->wrapExecutionType === WrapExecutionType::Disabled => WrapExecutionType::Disabled,
-            $value instanceof BaseDataCollectable && $this->wrapExecutionType === WrapExecutionType::TemporarilyDisabled => WrapExecutionType::Enabled,
-            default => throw new TypeError('Invalid wrap execution type')
-        };
-
         if ($value instanceof TransformableData && $this->transformValues) {
```
Nir-An, line 221: @rubenvanassche 
Please advise me if I really need to check the condition also for mapPropertyNames

docs/as-a-resource/transformers.md
```diff
@@ -72,7 +72,7 @@ The package will look through these global transformers and tries to find a suit
 
 ## Getting a data object without transforming
 
-It is possible to get an array representation of a data object without transforming the properties. This means `Carbon` objects won't be transformed into strings. And also, nested data objects and `DataCollection`s won't be transformed into arrays. You can do this by calling the `all` method on a data object like this:
+It is possible to get an array representation of a data object without transforming the properties. This means `Carbon` objects won't be transformed into strings. And also, nested data objects, `DataCollection`s won't be transformed into arrays and the names of the properties won't be change by `MapOutputName` . You can do this by calling the `all` method on a data object like this:
```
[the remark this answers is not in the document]
Nir-An: > Maybe we could add a section to the docs describing the functionality of `transform` for people who want a bit more control how to transform data transfer objects.

good idea, for now, I added a draft of how [disable the mapping](https://github.com/spatie/laravel-data/pull/199/files#diff-b4c931b0cca50f2922d28441675289a52f495fc860a305845987e87cd05a2c58R113)

src/Concerns/BaseDataCollectable.php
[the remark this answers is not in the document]
Nir-An: @rubenvanassche 
you're fine!! 

I did the changes right away after I saw your message, I asked just in case.
I already put the $mapPropertyNames after $wrapExecutionType 18 days ago
Please look at  [files changes](https://github.com/spatie/laravel-data/pull/199/files)

src/Transformers/DataTransformer.php
Nir-An: @rubenvanassche 

I think the PR is done, What do you think?  something is missing?

src/Contracts/TransformableData.php
Nir-An: already done

src/Concerns/TransformableData.php
Nir-An: mapPropertyNames with a false value

was already removed from the function

src/Transformers/DataTransformer.php
```diff
@@ -28,13 +28,15 @@ class DataTransformer
 
     public static function create(
         bool $transformValues,
+        bool $mapPropertyNames,
```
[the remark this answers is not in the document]
Nir-An: done!

Nir-An:
> Great PR! Some small remarks

 done.

src/Transformers/DataTransformer.php
rubenvanassche: Why did you remove the code block above and added it as `selectedBy` to WrapExecutionType?

src/Concerns/TransformableData.php
rubenvanassche: Shouldn't `mapPropertyNames: true` be set here if we want to keep the functionality of all as is?

Nir-An: > Shouldn't `mapPropertyNames: true` be set here if we want to keep the functionality of all as is?

yes it is. transform function has the mapPropertyNames true by default. 
https://github.com/Nir-An/laravel-data/blob/0a617c65a4bbf609b513bf457b3297d174d9b7f4/src/Concerns/BaseData.php#L100

src/Transformers/DataTransformer.php
Nir-An: the WrapExecutionType is charged to give us the type of wrap.
so I think it is the right place to make this decision, based on the instance of the value

but I can remove it if you want. 

source:
https://www.php.net/manual/en/language.types.enumerations.php ``do_stuff`` function

rubenvanassche: Let's keep it as is, I think such a refactor could be useful in the future when we need this functionality on multiple places. But this isn't the case right now so let's keep it close to where it is needed within the transformer.

Nir-An: > Let's keep it as is, I think such a refactor could be useful in the future when we need this functionality on multiple places. But this isn't the case right now so let's keep it close to where it is needed within the transformer.

I revert the changes regards to the wrap decision logic.
done.
thanks for your patience :+1:

rubenvanassche:
Thanks!

Nir-An:
> Thanks!

Thank you, your project is what makes Laravel more convenient and efficient to use, even more than it is today.

PR #200, opened 2022-08-23
Add support for timezone manipulation in casts and transformers

rubenvanassche:


francoism90:
@rubenvanassche This PR sounds interesting! :)

Would this mean we can use UTC (default) as (database) timezone and transformers to create dates for the front-end?

rubenvanassche:
That's correct!

francoism90:
@rubenvanassche That's great!

Do you have an example? Let's say the user-model has an attribute `timezone`, how could you load this into the dataobject?
Now it seems a bit hardcoded.

rubenvanassche:
At the moment you can't since attributes are quite dumb and don't know anything about the current state of the data object. I think dynamically adding casts using a new method would probably fix this but we don't need such feature right now. Always open for a PR adding this!

PR #207, opened 2022-08-30
Propagate generics to traits on collection classes

maxeckel:
Hey Folks,

using PHPStan/Larastan on level 9 currently reports issues regarding the generics on the collection classes.

For example the following code gives an error:

```php
/** @var DataCollection<int, TestData> $collection */
$collection = \App\DataObjects\TestData::collection([]);

$collection->each(function (\App\DataObjects\TestData $data) {
    echo $data->id;
});
```

PHPStan reports an error for the callback of the `each` method:

```
Parameter #1 $callback of method Spatie\LaravelData\DataCollection<int,App\DataObjects\TestData>::each() expects callable(mixed, (int|string)): mixed, Closure(App\DataObjects\TestData): void given.
```

This PR fixes these problems by annotating the used traits and by that propagate the types down to them.
It mimics how Laravel does the same thing for the Collection class.

Thank you for this and all the other great packages you provide!

### DIFF
--- a/src/CursorPaginatedDataCollection.php
+++ b/src/CursorPaginatedDataCollection.php
@@ -26,6 +26,8 @@ class CursorPaginatedDataCollection implements DataCollectable
     use IncludeableData;
     use WrappableData;
     use TransformableData;
+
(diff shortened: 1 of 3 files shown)


rubenvanassche:
Looking good!

PR #213, opened 2022-09-16
feat: Use generics with `Data::collection`

erikgaal:
As shown in the `types/Collection.php` tests, this PR ensures that the return types of `Data::collection` use generics.

Before, `SimpleData::collection($items)` would return a plain `DataCollection`.
Now, it returns `DataCollection<array-key, SimpleData>`.

Building upon #184, and makes sure that the correct generic types are returned when passing paginators.

```php
$users = User::paginate();
$collection = UserData::collection($users); // PaginatedDataCollection<array-key, UserData>

$users = User::cursorPaginate();
$collection = UserData::collection($users); // CursorPaginatedDataCollection<array-key, UserData>
```

This allows you to make stricter assertions in your project.

### DIFF
--- a/src/Contracts/BaseData.php
+++ b/src/Contracts/BaseData.php
@@ -13,29 +13,20 @@ use Spatie\LaravelData\DataPipeline;
 use Spatie\LaravelData\PaginatedDataCollection;
 
 /**
- * @template TValue of object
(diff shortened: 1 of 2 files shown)


rubenvanassche:
Cool, thanks!

PR #214, opened 2022-09-16
Improve pattern matching in DataCollectionAnnotationReader

erikgaal:
This PR fixes the `DataCollectionAnnotationReader` where it doesn't work with `@var` annotations that use the correct `DataCollection<TKey, TValue>` format.

I've also made the pattern matching a bit more readable by extracting repeated patterns and splitting it in two separate match calls.

### DIFF
--- a/src/Support/DataCollectionAnnotationReader.php
+++ b/src/Support/DataCollectionAnnotationReader.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\LaravelData\Support;
 
+use Illuminate\Support\Str;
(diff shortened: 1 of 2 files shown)


rubenvanassche:
Looking great, thanks!

rubenvanassche:
Hi @erikgaal, I had to reverse this PR because it broke some functionality. See the test suite for more details, I'v also enabled GH Action testing on PR's so we can be sure in the future all tests run. If you can fix this, I'll off course merge it again!

PR #215, opened 2022-09-19
fix: first and last page url are never null

erikgaal:
The `first_page_url` is always `string` for the SimplePaginator (1) and LengthAwarePaginator (2). `last_page_url` doesn't exist on `SimplePaginator`, but does exist on `LengthAwarePaginator` and is always `int`.

1: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Pagination/Paginator.php#L146
2: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Pagination/LengthAwarePaginator.php#L197


rubenvanassche:
Thanks!

PR #221, opened 2022-09-27
add ability to statically retrieve a data models rules

abenerd:
I sometimes only need the data models rules before working witht the model itself, this allows me to easily access those rules

### DIFF
--- a/src/Concerns/ValidateableData.php
+++ b/src/Concerns/ValidateableData.php
@@ -53,7 +53,7 @@ trait ValidateableData
         return;
     }
 
-    public static function getRules(string ...$fields): array
+    public static function getValidationRules(string ...$fields): array
     {
         $rules = app(DataValidatorResolver::class)
             ->execute(static::class, [])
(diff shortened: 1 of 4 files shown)


src/Concerns/ValidateableData.php
```diff
@@ -52,4 +52,17 @@ public static function withValidator(Validator $validator): void
     {
         return;
     }
+
```
rubenvanassche, line 55: Could you rename this to `getValidationRules`?

rubenvanassche (changes requested):
Small comment, could you also please add a small section to the docs, describing this feature? Might be handy for other people in the future!

abenerd:
Sure, I will do so tomorrow!

src/Concerns/ValidateableData.php
abenerd: Yeah, will also do that tomorrow with the documentation.

abenerd:
I added some documentation and updated the method name @rubenvanassche.

rubenvanassche:
Thanks!

PR #225, opened 2022-09-29
Improved pattern matching in DataCollectionAnnotationReader

erikgaal:
Second attempt at #214.

It's now backwards compatible with the generic notation without `TKey`.


erikgaal:
Hmm, looks like the Windows tests are failing due to difference in newline characters. But I'm not sure if and why my PR made a difference there.

rovansteen:
@rubenvanassche would you be able to include this one in a new release?

rubenvanassche:
Done

PR #226, opened 2022-09-29
Fix rules not being excluded

francoism90:
See https://github.com/spatie/laravel-data/discussions/223

I think you should include `array_key_exists($property->name, $overWrittenRules)`, otherwise all rules (including objects not part of the payload) will be included.


francoism90:
@rubenvanassche Could you please review this? :)

In the current implementation validation is broken, as it simple returns all rules (incl. sub) instead of checking the payload.

rubenvanassche:
See https://github.com/spatie/laravel-data/discussions/223#discussioncomment-3761115

PR #230, opened 2022-10-05
Add ExcludeWithout

francoism90:
@rubenvanassche This rule was missing, I've added the rule + adjusted docs for Laravel 9.x.

### DIFF
--- a/docs/advanced-usage/validation-attributes.md
+++ b/docs/advanced-usage/validation-attributes.md
@@ -37,7 +37,7 @@ class CustomRule extends ValidationRule
 
 ### Accepted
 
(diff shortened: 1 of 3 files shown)


rubenvanassche:
Looks like some tests are failing, could you please fix these?

francoism90:
@rubenvanassche I've fixed the date test, only the one I've added keeps failing. Could you please take a look? :)

I don't think you can use `Carbon::now()`, unless you specify a timezone? What are your thoughts?

Thanks!

rubenvanassche:
It has been fixed, thanks!

PR #231, opened 2022-10-09
improve getValidationRules to also retrieve wildcard rules

abenerd:
Previously wildcard rules such as `myarray.*` or `myarray.*.id` were discarded when retrieving the rules, this has been addressed so that all defined rules are not retrieved.


rubenvanassche:
Thanks!

PR #237, opened 2022-10-23
accept 'float' as data type

skyosev:
If you use custom creation method with *float* argument, it never gets called, because *gettype* returns 'double' for *float*  (see [official PHP docs](https://www.php.net/manual/en/function.gettype.php))


rubenvanassche:
Thanks!

PR #243, opened 2022-11-04
Fix typo in mime type validation rule

Sparclex:
Currently you get `BadMethodCallException: Method Illuminate\Validation\Validator::validateMimestypes does not exist.` when using the MimeTypes validation attribute since the method is called validateMimetypes without the additional "s". 

This PR fixes the typo


spekulatius (approved):

rubenvanassche:
Thanks!

PR #247, opened 2022-11-08
Add ability to modify properties before the data pipeline

ClintWinter:
Adds the ability for package users to inject logic into the process of their data object being created. More specifically, between normalization and the data pipeline.

The motivation for this is so that users may normalize the structure of their data. 

Personally, we've had situations where data coming in from an API has a different structure than data coming in from a csv import, but they both need to be massaged into the same DTO. 

This PR enables a natural way to do that *without having to override the magic creation methods*, which breaks all of the wonderful data pipeline functionality (mapping, casting, etc.) that makes this package so useful.


rubenvanassche:
Thanks!

PR #248, opened 2022-11-10
Add support for withoutTrashed to exists validation attribute

bentleyo:
This adds support for `withoutTrashed` to the `Exists` rule to bring it into line with the `Unique` rule.


rubenvanassche:
Thanks!

PR #252, opened 2022-11-15
[Docs] when should get Closure condition

Nir-An:
when on Lazy properties  should get Closure condition 

source :
https://github.com/spatie/laravel-data/blob/6001ec11f1ec841c43ac001c7f5dcdc09cbd6a1d/src/Lazy.php#L21
https://github.com/spatie/laravel-data/blob/69a3ec8a9e9a1b69fff58adde92bd9bd9bef0820/tests/DataTest.php#L260


rubenvanassche:
Thanks!

PR #281, opened 2022-12-01
ignore phpstorm attributes when instantiating and add readonly property

jhavenz:
This package throws an error while instantiating a data class if/when a built-in php storm attribute is used on a dto class.
e.g.:
```php
#[\JetBrains\PhpStorm\Immutable]
class PhpStormClassAttributeData extends Data
{
    public function __construct(public string $property)
    {
        $this->property = $property;
    }
}
```
and/or
```php
class PhpStormClassAttributeData extends Data
{
    #[\JetBrains\PhpStorm\Immutable]
    public string $property;

    public function __construct(string $property)
    {
        $this->property = $property;
    }
}
```
and/or
```php
class PhpStormClassAttributeData extends Data
{
    public function __construct(
       #[\JetBrains\PhpStorm\Immutable]
       public string $property
    ) {
       //
    }
}
```
_I'm using `\JetBrains\PhpStorm\Immutable` in these examples, but I've written this PR to ignore any of the other attributes that they have within the same namespace_

---

In the process of fixing this functionality, I came across the need to add the a 'isReadonly' property to the `DataType` class  ... I was actually surprised it wasn't already there. Though, despite me using this package (and source-diving) it quite a bit, maybe its elsewhere in the library and I'm just unaware. Just lmk.

Love the package, thx for the great work!


src/Support/DataClass.php
```diff
@@ -43,9 +44,9 @@ public function __construct(
 
     public static function create(ReflectionClass $class): self
     {
-        $attributes = collect($class->getAttributes())->map(
-            fn (ReflectionAttribute $reflectionAttribute) => $reflectionAttribute->newInstance()
-        );
+        $attributes = collect($class->getAttributes())
+            ->reject(fn (ReflectionAttribute $reflectionAttribute) => Str::startsWith($reflectionAttribute->getName(), 'JetBrains\\PhpStorm\\'))
```
rubenvanassche, line 48: Can't we just do a class_exists($reflectionAttribute->getName()) here? This would rule out all non existing attributes

src/Support/DataProperty.php
```diff
@@ -41,9 +42,9 @@ public static function create(
         ?NameMapper $classInputNameMapper = null,
         ?NameMapper $classOutputNameMapper = null,
     ): self {
-        $attributes = collect($property->getAttributes())->map(
-            fn (ReflectionAttribute $reflectionAttribute) => $reflectionAttribute->newInstance()
-        );
+        $attributes = collect($property->getAttributes())
+            ->reject(fn (ReflectionAttribute $reflectionAttribute) => Str::startsWith($reflectionAttribute->getName(), 'JetBrains\\PhpStorm\\'))
```
rubenvanassche, line 46: Same here

src/Support/DataType.php
```diff
@@ -22,6 +22,10 @@
 
 class DataType implements Countable
 {
+    private const READONLY_BITMASK = 128;
```
rubenvanassche, line 25: I don't think this is the right place to put readonly. A type is technically not readonly, a property is. So let's put it over there (DataProperty). Or am I missing something?

rubenvanassche (changes requested):
Thanks for your PR, I've left a few remarks + could you also provide tests for the readonly functionality. And is it possible to add support for PHP 8.2 readonly classes? So we can get it in one go!

Thanks!

jhavenz:
Ok, I've pushed up the changes I've made. If you decide you'd like to keep the `DataClass::isReadonly` property, this PR is good to go with the changes you've requested.

src/Support/DataClass.php
jhavenz: fixed

rubenvanassche:
Good point! Let's remove that property for now!

rubenvanassche:
Thanks!

PR #283, opened 2022-12-02
Always allow authorize to run

francoism90:
We are using Livewire, which doesn't use Request. It would be nice if the `authorize()` handle could always run, even when not using a request. :)

Other workarounds:
- Add `->through(CustomAuthorizedDataPipe::class)` to `pipeline()`
- Convert to request first, however this isn't really friendly as we're losing the types(?)

Think this is the easiest option. Let me know!

Thanks.


rubenvanassche:
I think a custom pipeline would be a better solution here. I think there are use cases where you don't want to do an authorize check when we're creating a data object from a model for example.

PR #291, opened 2022-12-07
Introduce support to return 201 status code in Json Responses

denjaland:
While refactoring our existing code from Laravel API resources to using Laravel Data, we noticed a lot of our tests were starting to fail, because we expect a `201 CREATED` response code when we are creating a new resource.

I checked the Laravel codebase, and they return a 201 when `$model->wasRecentlyCreated()` equals `true`.
Obviously, the same check can not be done in this package, but I believe it would make sense to test for the request method, and if it is a POST request, we can have return the Json response with a 201 status code, which is what I implemented in this PR.

Obviously, in order to ensure backwards compatibility, I disabled this by default by adding a new config parameter that is y default set to false, but can be enabled in the config file, so this can be release as a minor and not a major.

### DIFF
--- a/config/data.php
+++ b/config/data.php
@@ -67,4 +67,10 @@ return [
      * which will only enable the caster locally.
      */
     'var_dumper_caster_mode' => 'development',
+
+    /**
+     * When enabled, Laravel Data will return a JsonResponse with status 201 in stead of the
+     * default status code 200 when the toResponse() method is called in a request with method 'post'
+     */
+    'json_response_201_on_post_requests' => false
 ];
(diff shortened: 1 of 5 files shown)


rubenvanassche:
Hi @denjaland,

Thanks for the PR, I think this is more of a bugfix. So let's skip the whole config flag and do this by default! Could you make these changes? Thanks!

denjaland:
Will do.  I'm always careful making backwards incompatible changes, but I agree it's a fix.

denjaland:
@rubenvanassche updated PR

PR #294, opened 2022-12-09
Improve ide.json completion

bastien-phi:
This PR improves `ide.json` completion for 
- all parameters in `include`, `exclude`, `only`, `except`
- first parameter in `includeWhen`, `excludeWhen`, `onlyWhen`, `exceptWhen`

Capture d’écran du 2022-12-09 09-11-16

Capture d’écran du 2022-12-09 09-19-09


rubenvanassche:
Thanks!

PR #296, opened 2022-12-16
Allow sole method to be used on data collections without operator

bentleyo:
Laravel allows you to use `sole` without specifying an operator. It does this by checking the number of arguments.

This fixes an issue where the number of passed arguments was always 3.


rubenvanassche:
Thanks!

PR #303, opened 2022-12-22
Fill props from route parameters

ragulka:
This PR introduces a `FromRouteModel` attribute along with a `FillRouteModelPropertiesDataPipe`, based on the [discussion here](https://github.com/spatie/laravel-data/discussions/267#discussioncomment-4332848).

Note that this PR implements steps 1 & 2 from the comment linked above, and does not change the `Unique` attribute / rule. I felt this was outside of the scope of a single PR, and something that isn't strictly necessary from benefitting from route model binding.

@rubenvanassche would you mind reviewing this and let me know if I'm moving in the right direction here. I can try adding documentation as well once you feel the implementation and tests look good.

### Usage examples

This change allows filling data object properties from route model bindings. In the example below, if `SongData` is created from a request, the `id` property will be filled with `Song` model `$id`, _if_ the request contains a route model binding for the `Song` model.

```php
class SongData extends Data
{
    public function __construct(
        #[FromRouteModel('song')]
        public int $id,
        public string $name,
    ) {
    }
}
```

It's also possible to fill data properties from a different route model property. Here, `SongData::$name` will be filled from `Song::$title`.

```php
class SongData extends Data
{
    public function __construct(
        #[FromRouteModel('song')]
        public int $id,
        #[FromRouteModel('song', 'title')]        
        public string $name,
    ) {
    }
}
```

Finally, it's possible to control wether route model properties should replace existing properties - ie, for example if the request contains the model `$id` both in the body and in the route model. Replacements are turned on by default, but they can be turned off using the `replace: false` flag. Here, `SongData::$name` will be filled in from the route model _only_ if the request body does not contain the `name` key.

```php
class SongData extends Data
{
    public function __construct(
        #[FromRouteModel('song')]
        public int $id,
        #[FromRouteModel('song', replace: false)]        
        public string $name,
    ) {
    }
}
```

### Open questions and TODOs

- [ ] Documentation

**Should we support nested properties?**

Current implementation only supports shallow route model properties, ie, you cannot use nested keys:

```php
#[FromRouteModel('invoice', 'customer.name')]
```

To support the above, we'd probably need to convert the model to array, so we can fetch the nested property using `Arr::get()`

**Allow targeting the whole class with `FromRouteModel`?**

Perhaps it would be helpful to allow targeting the data class itself with this attribute. It would essentially mean that any data properties that can be mapped from the route model, would be mapped automatically. For example:

```php
#[FromRouteModel('song')]
class SongData extends Data
{
    public function __construct(
        public int $id, 
        public string $name,
    ) {
    }
}
```

In the example above, both id and name would be filled in from the route model, if present.


src/Attributes/FromRouteModel.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Spatie\LaravelData\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+class FromRouteModel
+{
+    public function __construct(
+        public string  $routeModel,
```
rubenvanassche, line 11: Let's call this `routeParameter` to be more in line with the router

src/Attributes/FromRouteModel.php
```diff
@@ -0,0 +1,16 @@
...
+        public ?string $routeModelProperty = null,
```
rubenvanassche, line 12: Let's call this modelProperty since it doesn't depend on the route

src/Attributes/FromRouteModel.php
```diff
@@ -0,0 +1,16 @@
...
+        public bool    $replace = true,
```
rubenvanassche, line 13: Let's call this `replaceWhenPresentInBody`

rubenvanassche (changes requested):
Cool PR!

Some really small naming changes and then this can be merged!

rubenvanassche:
You may also add the documentation like you described, so other people may also enjoy this feature!

As for nested properties, if it is not too much work and it can be coded cleanly in then I don't see a reason not to support it!

Targeting a whole class seems a bit overkill for me, I think this would be a bit too much.

Thanks!

ragulka:
@rubenvanassche thanks, I've updated the PR with your suggestions and also added support for nested route model properties.

However, there are a few things I wanted to discuss, before moving further.

Your comment on renaming `routeModel` to `routeParameter` made me think whether it would make sense to rename this whole feature to `FromRouteParameter`? It would make it a more generic approach to fill properties from any kind of route parameters, not just models. The data pipe doe snot really care, as long as the parameter is `Arrayable` (to support nested props).

As you [can see here](https://github.com/spatie/laravel-data/pull/303/commits/3d7d3fccdaeffe7aea0232a3b9a1e63a454c4503#diff-d2b7c22464a28eb815a7b3537f14b8e905d043df2c7c1d5cc93adbf014ebe365R29-R31), I am checking if the route parameter is an instance of `Model`, so that we can safely call `toArray()` on it. Perhaps we should check for `instanceof Arrayable`, or even `is_callable([$model, 'toArray'])`, instead?

I also realized during the tests that `toArray()` is called for every property that has the `FromRouteModel` attribute. Normally, `toArray()` should be rather quick, but I'm wondering if we should perhaps memoize the array version of the model/parameter, so that subsequent accesses would be quicker? Imagine a data object with 20 properties, where all of them would use this attribute - `toArray()` would be called 20 times.

What do you think?

bentleyo:
@ragulka instead of using `->toArray()` on the model and then `Arr::get` to get a property, would it be possible to use something like `data_get($model, $routeModelProperty)` instead?

ragulka:
> @ragulka instead of using `->toArray()` on the model and then `Arr::get` to get a property, would it be possible to use something like `data_get($model, $routeModelProperty)` instead?

@bentleyo that's a great idea. I'll try to refactor my PR around that. Thanks!

rubenvanassche:
Wow, this is looking great!

> ☝️ I was not entirely sure if that's the best approach. Maybe it should throw an exception instead?

Yeah let's do that

> ☝️ I'd like some feedback on how this is implemented. My reasoning to use a boolean false was that I didn't want to sacrifice the fallback to the data property name if the route parameter property name is not provided. However. I find using false to trigger the behavior above is somewhat ambiguous. Maybe flipping it so that if its set to a boolean true? Or perhaps introduce a different, optional param like useFullParameter?

An extra parameter seems indeed the best solution to me, because false doesn't really tell what's going to happen.

Would it be possible to also provide some docs? Then everyone can enjoy this lovely feature! I think this can even be a new docs page since it adds a lot of functionality.

The plan is to release v3 within a few hours, since this PR isn't breaking we can also add it in 3.1. I've already updated the branch to merge to.

ragulka:
> Wow, this is looking great!

@rubenvanassche Thanks 😊 

> > Maybe it should throw an exception instead?
> 
> Yeah let's do that

Perhaps there's a better way. Let me explain: Ideally, I'd like avoid throwing exceptions because filling from route parameters is sort-of optional anyway - properties will be filled if the route parameter exists. Also, there may be scenarios where it's more suitable to ignore the route parameter value if it's scalar, instead. With that in mind, I'll address the next point:

> An extra parameter seems indeed the best solution to me, because false doesn't really tell what's going to happen.

I agree that `false` doesn't tell whats going to happen. So I played around with the extra parameter a little and.. I think it's a bit awkward. What if I set both the property and the extra parameter at the same time?

```php
// What exactly is going on here??? 
#[FromRouteParameter('author', 'name', useFullRouteParameter: true )] 
public string $author
```
It's not at all clear what the intention of the developer is here. For the developer, it's unclear how the data property will be filled.

Given both points above, I'm leaning towards 2 separate attributes, with very clear purposes:

```php
// fill with the full route parameter / model
#[FromRouteParameter('author')] 
public AuthorData $author,

// fill with a scalar route parameter
#[FromRouteParameter('song-slug')] 
public string $slug,

// fill from a particular key/property of the route parameter,
// defaulting the key to the property name - in this case - `title`
#[FromRouteParameterKey('song')] 
public string $title,

// fill from a particular key/property of the route parameter using custom mapping
#[FromRouteParameterKey('song', 'name')]
public string $title
```

What do you think? `FromRouteParameterKey` could also be named `FromRouteParameterProperty`.

> Would it be possible to also provide some docs? Then everyone can enjoy this lovely feature! I think this can even be a new docs page since it adds a lot of functionality.

Of course. Just want to land on the final implementation first 😅

rubenvanassche:
Mhh yeah let's go with the two of them. It's a lot more readable.

> What do you think? FromRouteParameterKey could also be named FromRouteParameterProperty.

Yeah let's go with `FromRouteParameterProperty`, it makes it a lot more clear what's happening.

rubenvanassche:
Mhh GH closed that one automatically since it is not targetted to main.

ragulka:
@rubenvanassche that's fine, I can open a new one.

PR #304, opened 2022-12-22
Dynamic data classes

bentleyo:
@rubenvanassche this PR is a bit of a rough initial prototype, but I thought I'd demonstrate one way a collection of different data objects could potentially be implemented (see discussion #254).

This approach requires that the different data objects are of the same "type" and extend a base abstract class. That abstract class then chooses which data object implementation to use based on the provided data.

In the PR I've added an example of "Vehicle" as the base abstract class. Determining on the "type" property this data object will then be instantiated with the concrete implementation of either "Car" or "Horse", which have different properties, but still might be considered a "Vehicle" of sorts. Funny example I know 😂 

Since the child classes extend a base it's possible to add additional properties that will be shared by all, even if they're not required to determine the child class. I've provided an example of this with "passengers".

With this approach validation also works, even for collections and nested data. If the properties required to determine the child class aren't available then the validation rules are generated from the abstract class and validation errors will be generated from that.

If this is a feature you'd be interested in including I'm happy to make any changes you suggest or flesh out some tests. Also more than happy for you to improve upon this approach if you have any ideas! 😄

### DIFF
--- a/src/Concerns/DynamicDataTrait.php
+++ b/src/Concerns/DynamicDataTrait.php
@@ -36,6 +36,7 @@ trait DynamicDataTrait
         // Create the appropriate concrete data object from the properties
         /* @var class-string<DynamicData::class> $dataClass */
         $dataClass = static::dynamicClassName(Arr::only($properties, static::requiredPropertiesForResolving()));
+
         return $dataClass::from(...$payloads);
     }
 
--- a/src/Concerns/DynamicDataTrait.php
+++ b/src/Concerns/DynamicDataTrait.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Spatie\LaravelData\Concerns;
+
+use Illuminate\Support\Arr;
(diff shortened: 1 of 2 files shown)


rubenvanassche:
Hi @bentleyo,

Thanks for the PR, though it seems a really fun and cool feature I'm not going to merge it today. Because we first need to stabilize some other processes like validation and other parts of the package.

The good point is, if I look at your code that it is only a trait so people who want to use it can! Maybe in the future I'll review this again and add it in but first I want to make the package a bit more robust.

Thanks!

bentleyo:
Hi @rubenvanassche,

That's fair enough, it makes sense that stability and validation should be prioritised over features at the moment.

It's definitely a feature that has been really useful for our team, so I thought it would be worth sharing. The example @Richardggd mentioned in #254 is probably a more practical example of the same problem.

But you're right, we can absolutely just use the trait for now.

Cheers!

PR #307, opened 2022-12-22
Fetch data and parent fields for ide completion

bastien-phi:
IDE completion does not fetches parents fields for completion.

If we have 
```php
abstract class BaseSlot extends Data
{
    public function __construct(
        public CarbonImmutable $datetime,
        public ?int $project_activity_id,
    ) {
    }
}

class FinalSlot extends BaseSlot
{
    public function __construct(
        public ?bool $satisfied,
        CarbonImmutable $datetime,
        ?int $project_activity_id,
    ) {
        parent::__construct($datetime, $project_activity_id);
    }
}
```
completion provides only `satisfied` field : 
image

After this update, inherited fields are also completed : 
image

### DIFF
--- a/ide.json
+++ b/ide.json
@@ -5,7 +5,7 @@
             "complete": "classFields",
             "options": {
                 "fieldsFilter": {
-                    "fetch": "own",
+                    "fetch": "all",
                     "modifier": ["public"]
                 }
             },


rubenvanassche:
Thanks!

PR #310, opened 2022-12-29
Update normalizers.md

arcdigital:
changed four to five since there are five normalizers now.


rubenvanassche:
Good call, thanks!

Issue #309, opened 2022-12-29
Validation of optional nested Data attribute is impossible


denjaland:
Consider a `CreateAddressData` object defined as follows:

```php
class CreateAddressData extends Data {
  public function __construct(
    public readonly string $street,
    public readonly string $nbr,
    public readonly string $postal_code,
    public readonly string $city,
    public readonly string $cc
  ) {
  }

  public static function rules() {
    return [
      'street' => 'required|min:1|max:80',
      'nbr' => 'required|max:20',  
      'postal_code' => 'required|min:4|max:20',
      'city' => 'required|min:1|max:50',
      'cc' => 'required|size:2',
    ];
  }
}
```
And we have a `CreatePersonData` with the following definition:
```php
class CreatePersonData extends Data {
  public function __construct(
    public readonly string $first_name,
    public readonly string $last_name,
    public readonly string $email,
    public readonly ?CreateAddressData $address // Please note the address is OPTIONAL
  ) {
  }

  public static function rules() {
    return [
      'first_name' => 'required|min:1|max:50',
      'last_name' => 'required|min:1|max:50',
      'email' => 'required|email:rfc',
      'address' => '', // No validation rules for the address, as it's considered optional!
    ];
  }
```
The idea being that a person can optionally have an address, but when an address is given, the properties are correctly being validated...

The following code however throws a validation exception saying `The address.street field is required. (and 4 more errors)`:

```php
  CreatePersonData::validate(
    [
      'first_name' => $value,
      'last_name' => fake()->lastName(),
      'email' => fake()->email(),
    ]
  );
```

Shouldn't the nested data object only be validated when it is required in the containing data object or when it's optional and provided as part of the containing object?

I haven't checked the code yet (will do asap though), but Im only guessing you are merging the validation rules from the nested objects into the rules array for the containing one, but we should probably loop-validate instead, no?

Lovro1107:
Try adding `sometimes` rule to the `address` so it's validated when present.

https://laravel.com/docs/9.x/validation#validating-when-present

denjaland:
That doesn't work, because it would make the street optional; as soon as an address is provided, we need all address fields to be present.

I just went through the code, and just removing the required rule might fix it as it is automatically being added.  Will do some further testing.

denjaland:
hmm - removing the require on the address fields, now make the following test fail of course:

```php
test('it validates street', function (string $value) {
    CreateAddressData::validate(
        [
            'street' => $value,
[...]

PR #311, opened 2022-12-30
Fix https://github.com/spatie/laravel-data/issues/309

denjaland:
An attempt to fix the issue described in https://github.com/spatie/laravel-data/issues/309
I'm not really confident that I put the fix in the right place though; maybe it belongs inside the `DataClassValidationRulesResolver` instead.


rubenvanassche:
This will all be fixed with v3, gonna close this one for now.

PR #312, opened 2022-12-30
Handle custom regex validation rule

VGirol:
When using custom validation rules with regex and "|" character, we end up with this error message : 
preg_match(): No ending delimiter '/' found

in this example, each rule will fail :
```
use App\Support\DataTransfertObjects\Data;

class TestData extends Data
{
    public function __constructor(
        public readonly string $name
        public readonly string $other
    ) {}

    public static function rules()
    {
        return [
            'name' => 'required|regex:/test|ok/',
            'other' => ['required', 'regex:/test|ok/']
        ];
    }
}
```
This PR add support for regex rule with "|" character.
But there is a constraint : users must provide custom rules as array.
This is ok :
```
[
     'name' => ['required', 'regex:/test|ok/']
]
```
but not this :
```
[
     'name' => 'required|regex:/test|ok/'
]
```
That's why i added a note in the documentation.

### DIFF
--- a/docs/as-a-data-transfer-object/request-to-data-object.md
+++ b/docs/as-a-data-transfer-object/request-to-data-object.md
@@ -109,6 +109,9 @@ class SongData extends Data
 }
 ```
 
+> When using the regex / not_regex patterns, it is necessary to specify rules in an array instead of using "|" delimiters,
+> especially if the regular expression contains a "|" character.
+
 It is even possible to use the validationAttribute objects within the `rules` method:
 
 ```php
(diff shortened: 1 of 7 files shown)


rubenvanassche:
Hi @VGirol,

Would it be possible to update this branch for data v3? We're going to release this today and it has a completely rewritten validation logic.

VGirol:
Hi @rubenvanassche ,
I'll do it as soon as possible.

rubenvanassche:
Thanks, you'll probably also need to create a new pr since the branch which i retargetted it to was merged

PR #313, opened 2022-12-31
Making methods available in extending class

denjaland:
While I wanted to extend the class and implement the fix for #311 in my own implementation of the class (and DI'ing it), overriding the method, I noticed these functions were defined private, so I was unable access them in my class.


rubenvanassche:
Validation is completely rewritten in v3 so we won't need this anymore 😄

PR #329, opened 2023-01-20
Data v3

rubenvanassche:
This branch adds a completely rethought validation flow for Laravel Data.

TLDR:

Instead of creating all possible rules for nested data objects and collection we'll now generate rules based upon the payload provided. So if a nested data object for example is nullable and not provided, then all the nested rules for that data object are not added.

At this moment all tests are green! 🎉 

Currently not yet done (and possibly growing in the future):

- [x] Allowing to reference other fields of data objects, even when they are nested
- [x] Allow injecting values from data and route parameters
- [x] Allowing messages and attributes to be constructed for a validator recursively
- [x] Writing docs

This PR is breaking a lot, so this will probably be data v3.


francoism90:
@rubenvanassche Looks good! This would prevent the usage of `required_if` and `exclude_if`, etc. :)

PR #335, opened 2023-01-26
feat: extend the `make:data` command to be more flexible

samlev:
Adds some configuration for the `make:data` command, as well as the ability to override the defaults as required.

Documentation updates included.


rubenvanassche:
Thanks, I don't think we're going to make this command configurable by the config file, the options provided with the command seem enough to me. I'll update the code!

rubenvanassche:
In the end we decided to include it with #449, though not in the config by default. I've credited this PR in the readme.

PR #345, opened 2023-02-02
Add FieldReference to Before and After Validation Attribute

dtangdev:
```
    #[WithCast(DateTimeInterfaceCast::class)]
    #[Before(new FieldReference('end_at'))]
    public Carbon $start_at;

    #[WithCast(DateTimeInterfaceCast::class)]
    #[After(new FieldReference('start_at'))]
    public Carbon $end_at;
```

To prevent validation issues when the data is nested (inside a DataCollection for example)


rubenvanassche:
Thanks!

PR #348, opened 2023-02-04
Fix rules() typing

pyrou:
Since laravel-data v3, rules method expect a `ValidationContext` argument. Currently, laravel-data document the rules method as following on `ValidateableData` trait :

```php
/** 
 * @method static array rules(...$args)
 */
```

This lead to following static analyzer warning in our project :

> Parameter's name changed from 'args' to 'context'

This PR fixes the `@method` doc


rubenvanassche:
Thanks!

PR #349, opened 2023-02-07
Add WithCastable attribute

morrislaptop:
This PR adds a `WithCastable` attribute, inspired by https://laravel.com/docs/9.x/eloquent-mutators#castables

This will allow developers to create inline casters on their objects - avoiding the hassle of creating a specific caster classes.

This can be quite useful for small value objects, where you could write an object like

```php
<?php
namespace Spatie\LaravelData\Tests\Fakes\Castables;

use Spatie\LaravelData\Casts\Cast;
use Spatie\LaravelData\Casts\Castable;

class Email implements Castable
{
  public function __construct(public string $email) {

  }

  public static function castUsing(...$arguments)
  {
    return new class implements Cast {
        public function cast(DataProperty $property, mixed $value, array $context): mixed {
            return new Email($value);
        }
    };
  }
}
```

and use that within your data objects like

```php
class ForgotPasswordRequest extends Data
{
  public function __construct(
    #[WithCastable(Email::class)]
    public Email $email,
  ) {
  }
}
```

The DX could be improved (automatic detection of Castable on the property / skip having to specify the class to the attribute) however I thought it better to keep it light, in line with the `EnumCaster` approach.


morrislaptop:
Looks like [this commit](https://github.com/laravel/framework/commit/d50a2621e8ea9802d7aa4adaa07d856d3a634830) is breaking the tests for Laravel 10.x

rubenvanassche:
What if someone wants to combine an Eloquent and Data cast for one class? Won't this collide? Maybe prefixing the `castUsing` as `dataCastUsing`?

morrislaptop:
> dataCastUsing

morrislaptop:
> What if someone wants to combine an Eloquent and Data cast for one class? Won't this collide? Maybe prefixing the `castUsing` as `dataCastUsing`?

V. good point! Have updated.

rubenvanassche:
Thanks!

PR #354, opened 2023-02-11
Add toHtml for Livewire

francoism90:
This is my UserData class, using the introduced `LivewireLazy`:
```php
class UserData extends Data implements Wireable
{
    use WireableData;

    public function __construct(
        public Optional|string $id,
        public Optional|Lazy|string $email,
        #[DataCollectionOf(AddressData::class)]
        public Optional|Lazy|DataCollection $addresses,
    ) {
    }

    public static function fromModel(User $model): self
    {
        return new self(
            id: $model->getRouteKey(),
            email: Lazy::livewire(fn () => $model->email),
            addresses: Lazy::livewire(fn () => AddressData::collection($model->addresses)),
        );
    }
}
```

The `UserEditController.php` (basic example):
```php
<?php

class UserEditController
{
    public $user;
    
    public function mount()
    {
        $this->user = UserData::from(User::first())
            ->include('*');
    }
}
```

Without `LivewireLazy`:
```php
{{ $this->user->email }} // htmlspecialchars(): Argument #1 ($string) must be of type string, Spatie\LaravelData\Support\Lazy\DefaultLazy given
```

With `LivewireLazy` it works as expected:
```php
{{ $this->user->email }} // email@example.com
```

However the `AddressData` collection still results in a `LivewireLazy` collection instead of actually returning the data objects. I need to explicit call `{{ $this->user->addresses->resolve()->first() }}`.

Is this intended? Any workaround?

I think this PR would be a good fit, as Livewire is used to output the data object directly in most cases.

Thanks!

### DIFF
--- a/src/Lazy.php
+++ b/src/Lazy.php
@@ -7,6 +7,7 @@ use Illuminate\Database\Eloquent\Model;
 use Spatie\LaravelData\Support\Lazy\ConditionalLazy;
 use Spatie\LaravelData\Support\Lazy\DefaultLazy;
 use Spatie\LaravelData\Support\Lazy\InertiaLazy;
+use Spatie\LaravelData\Support\Lazy\LivewireLazy;
 use Spatie\LaravelData\Support\Lazy\RelationalLazy;
 
 abstract class Lazy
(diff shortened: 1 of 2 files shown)


francoism90:
It has been a week, could someone please take a look?

rubenvanassche:
Hi @francoism90,

I'm not going to merge this, since this is only a half solution to the problem. As soon as you're returning an array the whole thing will crash since you're required to return a string. Also chaining methods on properties is impossible.

It will work for strings, ints and floats but that's it.

We need a more structural solution to this problem, probably evaluating lazy when calling `include` but that's a major rewrite.

rubenvanassche:
I think this is fixed with #403, let's hope so 😄

PR #362, opened 2023-02-27
Add support for Validation Rule Contract in Rule rule

EriBloo:
Added support for `Illuminate\Contracts\Validation\ValidationRule` in `Spatie\LaravelData\Attributes\Validation\Rule`. Didn't remove depreciated contracts for compability reasons.


rubenvanassche:
Looks like some tests are failing

EriBloo:
Yes. Those failing tests are not related to my pull. After reverting my commit:

image

tests are still failing:

image

it seems they are related to typescript tranformer:

image

It would be weird if my commit, which only adds additional type, was breaking it.

EriBloo:
Ok I can see now that it fails in Laravel 9. That's to be expected since new Validation Rule didn't exist back then. Problem is that Laravel Data should have a new major version for Laravel 10 - the way it is right now either Laravel 9 or Laravel 10 is going to fail.

rubenvanassche:
Thanks!

PR #367, opened 2023-03-02
Quote non-standard object keys when transforming to Typescript

morphsteve:
Adds quotes to property names that contain special characters.

Rationale and further discussion may be found here: https://github.com/spatie/typescript-transformer/discussions/48


rubenvanassche:
That was quick, thanks!

PR #375, opened 2023-03-08
Allow laravel-data validation rules to be used in Laravel validator

hjanos1:
With the v2 version of laravel-data, it was possible to use all of the class-based validation rules in the `src/Attributes/Validation` folder with the Laravel validator, as an alternative to passing the rules by their string representation.
This provided a clean, typed representation of the rules in places, where using the rules as attributes would not work, like with Laravel Nova validation rules.

This use case has been, as of v3, broken, since [commit reference](https://github.com/spatie/laravel-data/commit/350d6fe61317910f7088042f787a1b72bd68c136#diff-e7432095734fd5269dcb9a94fd9d878cb6279dbcf736081dddd796554f0c7cd2).

This PR reintroduces support for using the validation rule classes provided by the package with the Laravel validator and adds a test to protect this use case from future breaks.

### DIFF
--- a/src/Attributes/Validation/StringValidationAttribute.php
+++ b/src/Attributes/Validation/StringValidationAttribute.php
@@ -2,18 +2,9 @@
 
 namespace Spatie\LaravelData\Attributes\Validation;
 
-use Spatie\LaravelData\Support\Validation\RuleDenormalizer;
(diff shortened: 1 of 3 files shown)


rubenvanassche:
Thanks!

PR #380, opened 2023-03-13
Introduce FormRequestNormalizer

bastien-phi:
This PR introduces a `FormRequestNormalizer` which allows to create a Data from validated content of a `FormRequest`. 

IMHO, it is more safe than to rely on the fact that a `Request` is `Arrayable` because unvalidated data won't be included in the `Data` with this new `FormRequestNormalizer`.

I did not include this new normalizer by default in the configuration because I think it could break existing applications if unvalidated data are used to create `Data` objects. I would be glad to enable it by default  if you think it's better.

### DIFF
--- a/config/data.php
+++ b/config/data.php
@@ -47,6 +47,7 @@ return [
      */
     'normalizers' => [
         Spatie\LaravelData\Normalizers\ModelNormalizer::class,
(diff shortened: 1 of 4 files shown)


rubenvanassche:
Thanks!

PR #381, opened 2023-03-13
Add Request validation data pipe

Yi-pixel:
Overwrite the data that needs to be verified in the request. You can add a `validationData` method like FromRequest to
provide the data used to participate in the form validation

- [framework/FormRequest.php at 10.x · laravel/framework](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Foundation/Http/FormRequest.php#L133)

### DIFF
--- a/docs/as-a-data-transfer-object/request-to-data-object.md
+++ b/docs/as-a-data-transfer-object/request-to-data-object.md
@@ -168,6 +168,34 @@ The rules will now look like this:
 ]
 ```
 
+### Overwrite Request Validation Data
+
+Overwrite the data that needs to be verified in the request. You can add a `validationData` method like FromRequest to
+provide the data used to participate in the form validation, but which must be a static method.
+
+A Request object will be provided as the first parameter.
+
+```php
+class SongData extends Data
+{
+    public function __construct(
+        public string $title,
+        public string $artist,
+    ) {
+    }
+    
+    public static function validationData(\Illuminate\Http\Request $request): array
+    {
+        return [
+            'title'  => $request->header('title'),
+            'artist' => $request->input('artist');
+        ];
+    }
+}
+
+
+```
+
 ### Rule attribute
 
 One special attribute is the `Rule` attribute. With it, you can write rules just like you would when creating a custom
--- a/src/Concerns/BaseData.php
+++ b/src/Concerns/BaseData.php
@@ -15,6 +15,7 @@ use Spatie\LaravelData\DataPipes\CastPropertiesDataPipe;
 use Spatie\LaravelData\DataPipes\DefaultValuesDataPipe;
 use Spatie\LaravelData\DataPipes\FillRouteParameterPropertiesDataPipe;
 use Spatie\LaravelData\DataPipes\MapPropertiesDataPipe;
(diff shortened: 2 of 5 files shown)


rubenvanassche:
Not sure if this is correct, since we use normalizers to convert objects into arrays of data. And I don't think we want to create a pipe which does the same because that would become confusing.

Maybe a beforeValidation method hook is a bit safer in this context where you pass in all the properties, you would lose the benefit of having a request object, but the whole system would be consistent which is always a win in my opinion.

Yi-pixel:
> Not sure if this is correct, since we use normalizers to convert objects into arrays of data. And I don't think we want to create a pipe which does the same because that would become confusing.
> 
> Maybe a beforeValidation method hook is a bit safer in this context where you pass in all the properties, you would lose the benefit of having a request object, but the whole system would be consistent which is always a win in my opinion.

It looks like Normalizers would be more appropriate 😇 .

This ensures that the data is already determined at the beginning.

Thanks

Yi-pixel:
@rubenvanassche 

Maybe there should be an optional [prepareForValidation](https://laravel.com/docs/10.x/validation#preparing-input-for-validation) static method on `Data` to do this (using Normalizers collection)?


In my example, this may require a new Normalizers, as I intend to use `Validation` from `laravel-data` entirely and not use `FormRequest` from `Laravel` anymore (the duplicate code is annoying).

eg:

```php
class Profile extends Controller
{
    public function update(UserData $data)
    {
        // TODO
    }
}
```

So I needed to organize my data before `Data` was created (and injected in the `Controller` method).

I noticed that when injecting directly with `Data` as a parameter, it always uses `$dataClass::from($container['request']` where `$container['request']` is not a `FormRequest`, so it shouldn't go to the existing ` FormRequestNormalizer`, so we also need a new Normalizers.

- [laravel-data/LaravelDataServiceProvider.php at 220f7f5acf2ea17efac1e9055722ccf55cbce011 · spatie/laravel-data](https://github.com/spatie/laravel-data/blob/220f7f5acf2ea17efac1e9055722ccf55cbce011/src/LaravelDataServiceProvider.php#L30)
- [laravel-data/FormRequestNormalizer.php at 220f7f5acf2ea17efac1e9055722ccf55cbce011 · spatie/laravel-data](https://github.com/spatie/laravel-data/blob/220f7f5acf2ea17efac1e9055722ccf55cbce011/src/Normalizers/FormRequestNormalizer.php#L11)

This new Normalizers should be behind the `FormRequestNormalizer`.

rubenvanassche:
> Maybe there should be an optional [prepareForValidation](https://laravel.com/docs/10.x/validation#preparing-input-for-validation) static method on Data to do this (using Normalizers collection)?

Mhh this would probably be something breaking, since Normalizers don't have any idea about what data object they are working for and you'll need that to call the method. But I'm not against adding such thing.

Maybe we should also change how the code works, at this moment when multiple payloads are provided to the from method of a data object. We run for each payload a normalizer + pipeline. It would make a lot more sense to run a normalizer on each payload merge them and then send them altogether through the pipeline.

We're working on a next major release of data but it will be a few months before we can release it. That being said, you're always welcome to send in a PR doing this to the `beyond-data` branch.

PR #385, opened 2023-03-21
Add support for models without all columns

martio:
Support for models without all columns.

Example:

```
$model = Team::query()
    ->with(relations: [
        'user:id,name',
        'club:id,name,country',
        'ranking:id,team_id,rank,rating',
     ])
     ->find(id: $id);
```

Currently throwing an error: `Illuminate\Database\Eloquent\MissingAttributeException: The attribute [created_at] either does not exist or was not retrieved for model [App\Models\Ranking]. in /var/www/html/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php:469` in `Laravel 10`.


rubenvanassche:
Thanks!

PR #399, opened 2023-04-01
Remove superfluous php version check

abenerd:
As the package requires `php:^8.1` we do not need to check for the version in the tests.


freekmurze:
Thanks!

PR #401, opened 2023-04-03
Improve performance when creating large amounts of Data objects

pjkellar:
# Description
When creating a large amount of Data objects in a single request cycle the repeated calls to `app()` that are used to resolve the dependencies for the `DataFromSomethingResolver` causes a performance bottleneck.

# Implementation

This PR registers the `DataFromSomthingResolver` as a singleton in the service provider and adds a `reset()` method on the class so that the internal state can be reset. This allows it to continue working when being called by either `from()` or  `withoutMagicalCreationFrom()`. This decreases the overall calls to the `app()` helper function (and subsequently reduces the downstream calls to other "magic methods" that resolve the dependancies) when instantiating a new `Data` object and increases the performance.
# Tests

Run the automated test suite and ensure that it completes successfully.
## Performance Tests

Using xdebug profiling and webgrind to to analyze the results I have seen an improvement in speed of approximately 12-18% when creating 30K instances of a Data object.  I have provided the Data object class `ZipCode` and customer caster class `ZipCaster`, as well as the sample dataset that was used for the testing. I have also provided a test script that was used to run the test 10 times and show the average.

[zips.txt.zip](https://github.com/spatie/laravel-data/files/11137440/zips.txt.zip)
[ziptest.php.zip](https://github.com/spatie/laravel-data/files/11137527/ziptest.php.zip)
[ZipCode.php.zip](https://github.com/spatie/laravel-data/files/11137531/ZipCode.php.zip)
[ZipCaster.php.zip](https://github.com/spatie/laravel-data/files/11137534/ZipCaster.php.zip)

Unzip and move the `zips.txt`.
Make sure the `ziptest.php` script has the right path to find the `zips.txt` file and run the script.
```php
php ziptest.php
```

### DIFF
--- a/src/Concerns/BaseData.php
+++ b/src/Concerns/BaseData.php
@@ -49,7 +49,7 @@ trait BaseData
 
     public static function from(mixed ...$payloads): static
     {
-        return app(DataFromSomethingResolver::class)->execute(
+        return app(DataFromSomethingResolver::class)->reset()->execute(
             static::class,
             ...$payloads
         );
@@ -57,7 +57,7 @@ trait BaseData
 
     public static function withoutMagicalCreationFrom(mixed ...$payloads): static
     {
-        return app(DataFromSomethingResolver::class)->withoutMagicalCreation()->execute(
+        return app(DataFromSomethingResolver::class)->reset()->withoutMagicalCreation()->execute(
             static::class,
             ...$payloads
         );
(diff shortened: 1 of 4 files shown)


rubenvanassche:
Hi @pjkellar,

Thanks for the PR, I really like the idea and it's a good PR. But adding these `reset` methods makes the code so complicated, and it just doesn't feel right.

We're working on a next major version of laravel-data, let's remove these two parameters from the constructor and add them to the execute method. Then the reset method isn't required anymore.

The problem is, it will take a few months before this next version will be released so for now I'm afraid you'll have to wait a bit or add this into your own project by extending some parts.

PR #408, opened 2023-04-07
feat: make `Lazy` macroable

innocenzi:
I'd like to add lazy and partial support for Hybridly, but instead of PRing built-in support in Laravel Data like Inertia, I'd like to implement it via a macro to keep the implementation flexible on my side.

```php
Lazy::macro('partial', function (\Closure $value): Partial {
    return hybridly()->partial($value);
});
```


rubenvanassche:
Thanks!

PR #419, opened 2023-04-18
Add merge method to Data Collections

didac-adria:
There is no merge method to Data Collections, and I think is quite useful. I'm adding it following the patters of the other methods.


rubenvanassche:
Thanks

PR #425, opened 2023-04-24
add ability to instantiate an eloquent cast when null value using null database values

daraghoshea:
Addresses situation when you may want the cast object to instantiated even when the database value is null.

The current `DataEloquentCast` get() method returns `null` if the database value is `null`.

This PR introduces a "nullable" [Cast Parameter](https://laravel.com/docs/eloquent-mutators#cast-parameters) that will ensure the Caster gets instantiated rather than return null. The nullable parameter was inspired by the [Spatie Laravel Enum package](https://github.com/spatie/laravel-enum#model-attribute-casting) that uses the same parameter, albeit for slightly different purposes.

The nullable parameter name can obviously be changed if preferred, but it was also applied to good effect in the [jessarcher/laravel-castable-data-transfer-object](https://github.com/jessarcher/laravel-castable-data-transfer-object/pull/13) package which extends the `spatie/data-transfer-object` package.

I am currently migrating from `spatie/data-transfer-object` to `spatie/laravel-data` on a number of projects. I have a few use cases where this would be incredibly useful, particularly when working with Livewire, and running into issues using @entangle not having any (default) values.

### DIFF
--- a/docs/advanced-usage/eloquent-casting.md
+++ b/docs/advanced-usage/eloquent-casting.md
@@ -74,3 +74,57 @@ Artist::create([
     ],
 ]);
 ```
+
+## Using defaults for null database values
+
+By default, if a database value is `null`, then the model attribute will also be `null`. However, sometimes you might want to instantiate the attribute with some default values.
+
+To achieve this, you may provide an additional `nullable` [Cast Parameter](https://laravel.com/docs/eloquent-mutators#cast-parameters) to ensure the caster gets instantiated.
+
+```php
+class Song extends Model
+{
+    protected $casts = [
+        'artist' => ArtistData::class . ':nullable',
+    ];
+}
+```
+
+This will ensure that the `ArtistData` caster is instantiated even when the `artist` attribute in the database is `null`.
+
+You may then specify some default values in the cast which will be used instead.
+
+```php
+class ArtistData extends Data 
+{
+    public string $name = 'Default name';
+}
+```
+
+```php
+Song::findOrFail($id)->artist->name; // 'Default name'
+```
+
+### Nullable collections
+
+You can also use the `nullable` argument in the case where you _always_ want a `DataCollection` to be returned.
+
+The first argument (after `:`) should always be the data class to be used with the `DataCollection`, but you can add `nullable` as a comma separated second argument.
+
+```php
+class Artist extends Model
+{
+    protected $casts = [
+        'songs' => DataCollection::class.':'.SongData::class.',nullable',
+    ];
+}
+```
+
+```php
+$artist = Artist::create([
+    'songs' => null
+]);
+
+$artist->songs;             // DataCollection
+$artist->songs->count();    // 0
+```
--- a/src/Concerns/TransformableData.php
+++ b/src/Concerns/TransformableData.php
@@ -28,6 +28,6 @@ trait TransformableData
 
     public static function castUsing(array $arguments)
     {
-        return new DataEloquentCast(static::class);
+        return new DataEloquentCast(static::class, $arguments);
     }
 }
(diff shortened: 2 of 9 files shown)


linaspasv:
Wow, this is a nice PR. This is what I was just looking for myself. I was trying to use Data objects to initialize JSON fields with default values typed in Data object if there is nothing set in the database.

linaspasv:
> Not entirely sure if nullable is the correct name for the parameter here. Adding such a parameter gives me the idea that the value can be a data object or null. Now the cast is nullable by default, maybe that wasn't the best design idea in hindsight, maybe renaming the thing to `emptyable`?
> 
> Not a big fan of that name so if anyone has a better proposal, please react 😄

By the way, I was just wondering why for NULL we also return NULL instead of an empty Data object with default values. I do very often use Data casts for ``nullable()`` json fields to cast them with default values but in order for this to work I need to override ``DataEloquentCast``.

```php
class DataEloquentCast extends BaseDataEloquentCast
{
    /**
     * @throws
     */
    public function get($model, string $key, $value, array $attributes): ?BaseData
    {
        if ($value === null) {
            $value = '{}';
        }

        return parent::get($model, $key, $value, $attributes);
    }
}
```

Maybe if we could have this behavior as a default and provide an easy flag for the user to determine if Data cast was created from a nullable field (on the other hand there is getOriginal() where user can check if it's null if needed), we would not even need ``nullable`` keyword.

Finally, this PR brings something that I would use very often, just would prefer this behavior to be a default one. :-)

daraghoshea:
> Is it possible with your PR to load default values defined in the Data object? I have a nested data and it did not work for me. I have ended doing this.
> 
> ```
>     public function get($model, string $key, $value, array $attributes): ?BaseData
>     {
>         if ($value === null) {
>             $payload = ($this->dataClass)::empty();
> 
>             return ($this->dataClass)::from($payload);
>         }
> 
>         return parent::get($model, $key, $value, $attributes);
>     }
>     
> ```

I had not looked at nested data in tests. I'll try find time this week, but my guess is that it should work. Not sure why it didn't for you.

linaspasv:
> 

After 3.5.0 changes for default values it actually did start to work.

I just override DataEloquentCast with this and it works perfectly now:

```php
    public function get($model, string $key, $value, array $attributes): ?BaseData
    {
        if ($value === null) {
            $value = '{}';
        }

        return parent::get($model, $key, $value, $attributes);
    }
   ```

**Comment by daraghoshea** [comment]
> > 
> 
> After 3.5.0 changes for default values it actually did start to work.
> 
> I just override DataEloquentCast with this and it works perfectly now:
> 
> ```
>     public function get($model, string $key, $value, array $attributes): ?BaseData
>     {
>         if ($value === null) {
>             $value = '{}';
>         }
> 
>         return parent::get($model, $key, $value, $attributes);
>     }
> ```

Yes, that's how this PR handles it also.

```php
    public function get($model, string $key, $value, array $attributes): ?BaseData
    {
        if (is_null($value) && in_array('nullable', $this->arguments)) {
            $value = '{}';
        }

        if ($value === null) {
            return null;
        }

        $payload = json_decode($value, true, flags: JSON_THROW_ON_ERROR);

        return ($this->dataClass)::from($payload);
    }
```

linaspasv:
> 

Yep, I took inspiration from your PR, just it was not working before 3.5.0 as there were no way to define default values for a non-generic properties, so needed to do some ugly things to make it work before. :-)

rubenvanassche:
> By the way, I was just wondering why for NULL we also return NULL instead of an empty Data object with default values. I do very often use Data casts for nullable() json fields to cast them with default values but in order for this to work I need to override DataEloquentCast.

When something is nullable, I expect the object or null out of it when empty/non existing. Returning the object or a default object in that way just feels wrong.

 We'll add the `:default` keyword looks correct to me 😄 

Thanks!

PR #433, opened 2023-05-03
Add missing constructor parameters to error message

Sparclex:
Currently only the given parameters are displayed when failing to create a data object. But if a Data Object has a lot of parameters it is hard to find the missing property. 

This PR suggests displaying the missing constructor property names as well in the error message.


rubenvanassche:
Thanks!

Sparclex:
> I was on the verge of PRing this. This would be very welcome, and I even suggest swapping "parameters given" and "parameters missing" so the latter becomes first. We could go as far as removing "parameters given" maybe.

Yes I think changing the order would be even better. But I would not remove "parameters given" since it already helped me sometimes, when I forgot to add `MapName` for example.

PR #444, opened 2023-05-15
Replace enum package example with native enum

Kehet:
This took me quite a while before I understood to read the source code

Relevant code https://github.com/spatie/laravel-data/blob/5e170162c573085b7529cd1cd8881201f22b5110/src/Casts/EnumCast.php#L12-L15

### DIFF
--- a/docs/as-a-data-transfer-object/casts.md
+++ b/docs/as-a-data-transfer-object/casts.md
@@ -18,15 +18,13 @@ class SongData extends Data
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Kehet:
Actually this seems to be bigger problem with this page since spatie/enum package isn't supported .... I'll write some example caster for that.

Kehet:
I decided to split this example to another PR #445 in case it is not needed

flexchar:
Would love to see this, as Spatie also serves as a great example on "how to" in the Laravel/PHP world :)

freekmurze:
Thank you!

PR #445, opened 2023-05-15
Another example for custom casters page

Kehet:
This adds example how to use spatie/enum package. Related to #444


rubenvanassche:
Thanks for the example, but since PHP and laravel-data now has native enums I think we want to phase out these Spatie enums. So I don't think adding an example how to write an enum cast would be beneficial to this transition.

If you have an idea for a cast which does something else than Spatie enums I'll gladly accept it.

PR #448, opened 2023-05-17
Changing Request to FormRequest

mikkolukas:
This small detail threw me off into a rabbit hole for hours, trying to extend `Illuminate\Http\Request` 😵

By letting the code example extend from `FormRequest` instead, it is made very explicit how important this small detail is.


rubenvanassche:
Thanks!

PR #449, opened 2023-05-17
[fix] move make:data command options to config

pepperfm:
This feature should make the creation of data objects classes more comfortable.
It helps to avoid long spamming `--namespace="..."` option.


rubenvanassche:
Thanks, I'm going to merge this but rework it a little bit, I want to keep the configuration as clean as possible so we're going to document it but not add it to the config file itself.

PR #451, opened 2023-05-19
Change required rule inferrer to be false when array has present rule

falko100:


falko100:
This is not working yet. 

```php
#[Present]
#[ArrayType]
public readonly array $someIds,
```
This does not get the rules in the inferrer.

PR #497, opened 2023-07-07
Fixed target namespace when creating files with Laravel Idea

andrey-helldar:
If you specify a target directory, such as "app/Data", the files will always be created in this directory, even if the main namespace of the application or package is in another folder, such as "src":

```
app
  Data
    SomeData
src
  Console
  Enums
```

Replacing the target directory with a namespace parameter will allow the Laravel Idea plugin to create files inside the application's Namespace:

```
src
  Console
  Data
    SomeData
  Enums
```

Untitled Project

### DIFF
--- a/ide.json
+++ b/ide.json
@@ -41,7 +41,7 @@
             "classSuffix": "Data",
             "files": [
                 {
-                    "directory": "/app/Data",
+                    "appNamespace": "Data",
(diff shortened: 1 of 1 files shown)


rubenvanassche:
Wow this is great, thanks!

PR #505, opened 2023-07-18
Adding Hidden Attribute

dbpolito:


rubenvanassche:
Could you please revert to the original version of your PR?

In contraire to what the user in #472 was saying, we don't have a feature that will hide properties from payloads but will include them when saving to the database.

Now we're in an inconsistent state where some properties using Hide will be saved to the database and others except and only wont. If we want to include properties when saving to the database a manual transform should be executed and a feature should be implemented there which enables or disables excepts(basically the Hide of this PR) and only's. But such a feature is more something for v4 which totally rethinks how transformation contexts work.

dbpolito:
@rubenvanassche Done 👍

rubenvanassche:
Thanks!

linaspasv:
> Could you please revert to the original version of your PR?
> 
> In contraire to what the user in #472 was saying, we don't have a feature that will hide properties from payloads but will include them when saving to the database.
> 
> Now we're in an inconsistent state where some properties using Hide will be saved to the database and others except and only wont. If we want to include properties when saving to the database a manual transform should be executed and a feature should be implemented there which enables or disables excepts(basically the Hide of this PR) and only's. But such a feature is more something for v4 which totally rethinks how transformation contexts work.

Do we have anything in v4 that would allow us to save to Database but keep sensitive data from output?

PR #507, opened 2023-07-21
Allowing collection to be created passing null

dbpolito:
Currently if you call `Data::collection(null)`, it breaks... it's useful for when you are doing something like `Data::collection($var?->collection)` and avoiding doing and if or `Data::collection($var?->collection ?? [])`...

Also keeps consistent with Laravel's Collection, which you can create passing null and will just create an empty collection.


rubenvanassche:
Thanks!

PR #508, opened 2023-07-21
Update quickstart.md

afpinedac:
Removed unused constructor parameter.
The example shows that the constructor receives 4 parameters but it was sending 5 arguments.


rubenvanassche:
Thanks!

PR #510, opened 2023-07-25
Added Ulid validation rule.

oreillysean:
This commit has
 
* an associated test
* all tests passing
* updated the docs
* ulid validation in laravel is documented here [https://laravel.com/docs/9.x/validation#rule-ulid](https://laravel.com/docs/9.x/validation#rule-ulid)


rubenvanassche:
Thanks!

PR #526, opened 2023-08-08
Add abstract eloquent casts

rubenvanassche:
Sometimes you have an abstract parent data object with multiple child data objects, for example:

```php
abstract class RecordConfig extends Data
{
    public function __construct(
        public int $tracks,
    ) {}
}

class CdRecordConfig extends RecordConfig
{
    public function __construct(
        int $tracks
        public int $bytes,
    ) {
        parent::__construct($tracks);
    }
}

class VinylRecord extends RecordConfig
{
    public function __construct(
        int $tracks
        public int $rpm,
    ) {
        parent::__construct($tracks);
    }
}
```

A model can have a JSON field which is either one of these data objects:

```php
class Record extends Model
{
    protected $casts = [
        'config' => RecordConfig::class,
    ];
}
```

You can then store either a `CdRecordConfig` or a `VinylRecord` in the `config` field:

```php
$cdRecord = Record::create([
    'config' => new CdRecordConfig(tracks: 12, bytes: 1000),
]);

$vinylRecord = Record::create([
    'config' => new VinylRecord(tracks: 12, rpm: 33),
]);

$cdRecord->config; // CdRecordConfig object
$vinylRecord->config; // VinylRecord object
```

When a data object class is abstract and used as an Eloquent cast then this feature will work out of the box.

The child data object value of the model will be stored in the database as a JSON string with the class name as the key:

```json
{
    "type": "\\App\\Data\\CdRecordConfig",
    "value": {
        "tracks": 12,
        "bytes": 1000
    }
}
```

When retrieving the model, the data object will be instantiated based on the `type` key in the JSON string.

# Abstract data class morphs

By default, the `type` key in the JSON string will be the fully qualified class name of the child data object. This can break your application quite easily when you refactor your code. To prevent this, you can add a morph map like with [Eloquent models](https://laravel.com/docs/10.x/eloquent-relationships#polymorphic-relationships). Within your `AppServiceProvivder` you can add the following mapping:

```php
use Spatie\LaravelData\Support\DataConfig;

app(DataConfig::class)->enforceMorphMap([
    'cd_record_config' => CdRecordConfig::class,
    'vinyl_record_config' => VinylRecordConfig::class,
]);
```

# Why no DataCollection support?

At this point in type a `DataCollection` can only be of one type only, so no use to make a cast for abstract child classes for now.

### DIFF
--- a/docs/advanced-usage/eloquent-casting.md
+++ b/docs/advanced-usage/eloquent-casting.md
@@ -62,7 +62,7 @@ class CdRecordConfig extends RecordConfig
     }
 }
 
-class VinylRecord extends RecordConfig
+class VinylRecordConfig extends RecordConfig
 {
     public function __construct(
         int $tracks
@@ -92,11 +92,11 @@ $cdRecord = Record::create([
 ]);
 
 $vinylRecord = Record::create([
-    'config' => new VinylRecord(tracks: 12, rpm: 33),
+    'config' => new VinylRecordConfig(tracks: 12, rpm: 33),
 ]);
 
 $cdRecord->config; // CdRecordConfig object
-$vinylRecord->config; // VinylRecord object
+$vinylRecord->config; // VinylRecordConfig object
 ```
 
 When a data object class is abstract and used as an Eloquent cast then this feature will work out of the box.


bentleyo:
@rubenvanassche this is really cool, I like the approach you took!

Is the plan to extend this functionality in the future so it can be used with collections etc.?

Just checking as we're currently [using a trait to handle some dynamic data class functionality](https://github.com/spatie/laravel-data/pull/304) and it would be great if there was built-in support for this in the package!

rubenvanassche:
I really want to implement such a thing, but am not sure at the moment how we're going to do it. Especially since validation of such collections for example will not work due to the different types. 

This was really a simple fix for a problem we we're having in one of ours projects.

But as soon as v4 is out I'll start taking a look at these types of collections.

bentleyo:
Thanks for replying @rubenvanassche! In our situation it was fine for us to have a property on the abstract class that identified which concrete version of the class to use (and that's the approach we took with the trait I linked above), but I can see how that might not be a solution everyone would be happy with.

In some ways I feel like having a piece of data on the abstract data class and having it determine which concrete version of the class to validate against / instantiate makes sense. Without the identifying data in the abstract class it was really awkward to work with because we'd always need to pass in the type / class name of the data object to use externally - especially awkward when working with collections of the same abstract type.

There's probably better ways to solve that issue, but for us it felt like a reasonable trade-off requiring some piece of variant-identifying data in the objects themselves.

Cheers mate!

Issue #500, opened 2023-07-13
Data collection properties that are optional do not have their rules used


KaloyanYosifov:
**✏️ Describe the bug**
Optional Data Collection property's rules are not taken into consideration when it's not part of the request body. 
I am trying to create a DTO that has to have one of the collections present if the other does not. (`someData` is required if `someOtherData` is not present. And vice versa). 

The problem is that if none of the properties are passed to the DTO, then it does not consider the rules of those optional properties.

**↪️ To Reproduce**
```php

it('does not return validation rules for optional collections', function () {
    class ChildData extends Data
    {
        public function __construct(
            public string $key,
            public string $name,
        ) {

        }
    }

    class BaseData extends Data
    {
        public function __construct(
            public int $id,
            public string $name,
            #[RequiredWithout('someOtherData')]
            #[\Spatie\LaravelData\Attributes\DataCollectionOf(ChildData::class)]
            public \Spatie\LaravelData\DataCollection|Optional $someData,
            #[RequiredWithout('someData')]
            #[\Spatie\LaravelData\Attributes\DataCollectionOf(ChildData::class)]
            public \Spatie\LaravelData\DataCollection|Optional $someOtherData,
        ) {

        }
    }

    // Validation rules for someData and someOtherData not present
    dd(BaseData::getValidationRules(['name' => 'test', 'id' => 1]));
});
```

**✅ Expected behavior**
Validation rules are always present, so we can have cases like `requiredWithout`. 

**🖥️ Versions**

Laravel: 9.52.10
Laravel Data: 3.7.0
PHP: 8.1

---

It seems to be because of the condition on `DataValidationRulesResolver::78` where if the property is optional and the field is not in the payload we skip validation. But in this case, I want the two fields to be optional but throw a validation error that at least one of them has to be set. 

I am happy to do a PR if you think we could add a check for explicit `Required` or `Exclude` rules and allow validation always. Or let me know if I am missing something

rubenvanassche:
> I am happy to do a PR if you think we could add a check for explicit Required or Exclude rules and allow validation always. Or let me know if I am missing something

You're always welcome to do such a thing, I certainly would accept a well tested PR. Good luck!

rubenvanassche:
@KaloyanYosifov, something to consider: https://github.com/spatie/laravel-data/issues/500

We basically have the same problem with data objects and a null value.

rubenvanassche:
Fixed thanks to @KaloyanYosifov!

PR #532, opened 2023-08-14
Fix for no rules created on optional|nullable Data object and Collection

KaloyanYosifov:
This fixes the issue described on #500


rubenvanassche:
Looking great, thanks!

PR #533, opened 2023-08-15
Add missing Typehint

Enaah:
Add the Typehint in the example as in the Interface Class (`src/Casts/Castable.php`)


rubenvanassche:
Thanks!

PR #534, opened 2023-08-17
Update type

markhuot:
It should be possible to re-map a collection of models to a collection of arrays with something like

```php
$collection->map(fn (Model $item) => ['id' => $item->id])
```

But the current type hit expects `->map()` to return the same typed collection as the input. The updated type infers the returned type from the closure return.

This is the same type definition that Laravel's collection `->map()` uses, https://github.com/laravel/framework/blob/85ad68cb0bcb5143e0e915098097c6d23ea22e7e/src/Illuminate/Collections/Collection.php#L767-L774

### DIFF
--- a/src/Concerns/EnumerableMethods.php
+++ b/src/Concerns/EnumerableMethods.php
@@ -29,9 +29,11 @@ trait EnumerableMethods
     }
 
     /**
-     * @param callable(TValue, TKey): TValue $map
+     * @template TMapValue
      *
-     * @return static
+     * @param  callable(TValue, TKey): TMapValue  $callback
+     *
+     * @return static<TKey, TMapValue>
      */
     public function map(callable $map): static
     {


markhuot:
Note: it doesn't look like there are any tests for types so I didn't add anything explicitly for this. Let me know if I should start a test file for types to handle changes like this.

rubenvanassche:
I get your point but this is done with a very specific reason, these are DataCollections and they should 🥁 contain Data objects. So mapping them to something else than data would not be correct to the DataCollection type, especially all the extra logic on these collections which expect Data objects.

In this case I would first get the collection by calling `toCollection()` on it and then map.

markhuot:
Ah! That makes sense calling `->toCollection()` to first get an abstracted collection. Thanks for the tip.

PR #546, opened 2023-09-04
doc update regarding validation rules on object creation

splatEric:
Super helpful package, thanks a lot!!

But having just tripped over this [issue](https://github.com/spatie/laravel-data/discussions/357), I thought it might be helpful to suggest an addition to the documentation for future users.

### DIFF
--- a/docs/advanced-usage/validation-attributes.md
+++ b/docs/advanced-usage/validation-attributes.md
@@ -18,6 +18,8 @@ class SongData extends Data
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rubenvanassche:
We already have a section on this: [We're reworking validation completely for v4, this will also include new docs specific on validation. I think](https://spatie.be/docs/laravel-data/v3/as-a-data-transfer-object/request-to-data-object#content-validating-a-data-object-without-request)

splatEric:
okay. seems a shame not to be willing to put in an extra bit of information that could save users some time who aren't aware of it. But it's your project :shrug:

PR #569, opened 2023-10-01
Fix ipv4 & ipv6 links in docs

Klaas058:
The ipv4 & ipv6 anchor tag in the links to the Laravel validation docs were not right


freekmurze:
Thanks!

Issue #571, opened 2023-10-02
Add missing & update existing validation attributes


Klaas058:
### Discussed in https://github.com/spatie/laravel-data/discussions/570

<div type='discussions-op-text'>

<sup>Originally posted by **Klaas058** October  1, 2023</sup>
I cross-referenced the [validation attributes](https://github.com/spatie/laravel-data/blob/main/docs/advanced-usage/validation-attributes.md) with the [Laravel 9.x available validation rules](https://laravel.com/docs/9.x/validation#available-validation-rules) and found that there are 18 validation rules that are not implemented as validation attributes yet.

I couldn't find information about which Laravel version this package needs to support, so depending on the version support, I would like to add the following:

* Edit I found the requirements docs, it says [9 or higher](https://spatie.be/docs/laravel-data/v3/requirements) so I will only create PR's for version 8.* validators for now

- [ ] declined_if [8.71.0](https://github.com/laravel/framework/releases/tag/v8.71.0)
- [ ] declined [8.71.0](https://github.com/laravel/framework/releases/tag/v8.71.0)
- [ ] mac [8.77.0](https://github.com/laravel/framework/releases/tag/v8.77.0)
- [ ] required_array_keys [8.82.0](https://github.com/laravel/framework/releases/tag/v8.82.0)
- [ ] exclude_with [9.6.0](https://github.com/laravel/framework/releases/tag/v9.6.0)
- [ ] doesnt_start_with [9.17.0](https://github.com/laravel/framework/releases/tag/v9.17.0)
- [ ] doesnt_end_with [9.24.0](https://github.com/laravel/framework/releases/tag/v9.24.0)
- [ ] min_digits [9.26.0](https://github.com/laravel/framework/releases/tag/v9.26.0)
- [ ] max_digits [9.26.0](https://github.com/laravel/framework/releases/tag/v9.26.0)
- [ ] uppercase [9.40.0](https://github.com/laravel/framework/releases/tag/v9.40.0)
- [ ] lowercase [9.40.0](https://github.com/laravel/framework/releases/tag/v9.40.0)
- [ ] ascii [9.44.0](https://github.com/laravel/framework/releases/tag/v9.44.0)
- [ ] decimal [9.45.0](https://github.com/laravel/framework/releases/tag/v9.45.0)
- [ ] missing_with_all [9.49.0](https://github.com/laravel/framework/releases/tag/v9.49.0)
[...]

PR #572, opened 2023-10-02
Add Declined and DeclinedIf validation attributes

Klaas058:
Adds the **Declined** and the **DeclinedIf** if validation rules.

The implementation strategy is equal to the [Accepted](https://github.com/spatie/laravel-data/blob/main/src/Attributes/Validation/Accepted.php) and [AcceptedIf](https://github.com/spatie/laravel-data/blob/main/src/Attributes/Validation/AcceptedIf.php).

As part of #571


rubenvanassche:
Thanks!

PR #577, opened 2023-10-05
feat: support `Arrayable` when casting to `DataCollection`

innocenzi:
This pull request adds support for casting `Collection` instances to `DataCollection`. 

I had to use `->toArray` in multiple scenarios when using `DataCollection` casts, but I think that should be handled by `laravel-data` directly.


rubenvanassche:
Thanks!

PR #581, opened 2023-10-09
Fetch attributes from parent classes to allow reusability

bastien-phi:
This PR changes class attributes resolution fetching parent classes attributes as well as current class attributes.

This allows the developer to factorize attributes declarations parent classes.

For example if we want every data to be mapped with `SnakeCaseMapper`, we currently need to declare `#[MapName(SnakeCaseMapper::class)]` on every single `Data` class.

With this PR, we can declare a `BaseData` with the mapper and every child class with inherit with the mapper.

```php
#[MapName(SnakeCaseMapper::class)]
abstract class BaseData extends Data {}

class MyData extends BaseData {
    public function __construct(
        public string $firstName,
    ) {}
}

MyData::from(['first_name' => 'John']); // work as expected
```

In case the attribute is redeclared in child class, the child attribute has higher priority (parents attributes are push at the end of the collection).

### DIFF
--- a/src/Support/DataClass.php
+++ b/src/Support/DataClass.php
@@ -15,6 +15,7 @@ use Spatie\LaravelData\Contracts\ResponsableData;
 use Spatie\LaravelData\Contracts\TransformableData;
 use Spatie\LaravelData\Contracts\ValidateableData;
 use Spatie\LaravelData\Contracts\WrappableData;
+use Spatie\LaravelData\Data;
 use Spatie\LaravelData\Mappers\ProvidedNameMapper;
 use Spatie\LaravelData\Resolvers\NameMappersResolver;
 use Spatie\LaravelData\Support\Lazy\CachedLazy;
@@ -49,9 +50,7 @@ class DataClass
 
     public static function create(ReflectionClass $class): self
     {
-        $attributes = collect($class->getAttributes())
-            ->filter(fn (ReflectionAttribute $reflectionAttribute) => class_exists($reflectionAttribute->getName()))
-            ->map(fn (ReflectionAttribute $reflectionAttribute) => $reflectionAttribute->newInstance());
+        $attributes = static::resolveAttributes($class);
 
         $methods = collect($class->getMethods());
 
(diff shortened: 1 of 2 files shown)


tests/Support/DataClassTest.php
```diff
@@ -120,3 +136,18 @@ public static function fromDummyModel(DummyModel $model)
...
+{
+}
+
+class ChildData extends ParentData
+{
+    public function __construct(
+        public DateTimeInterface $dateTime
+    ) {
+    }
+}
```
rubenvanassche, line 153: Can you move this inside the test and give it more test specific names like TestRecursiveAttributesChildData?

src/Support/DataClass.php
```diff
@@ -81,6 +80,21 @@ public static function create(ReflectionClass $class): self
...
     }
 
+    protected static function resolveAttributes(
+        ReflectionClass $class
+    ): Collection {
+        $attributes = collect($class->getAttributes())
+            ->filter(fn (ReflectionAttribute $reflectionAttribute) => class_exists($reflectionAttribute->getName()))
+            ->map(fn (ReflectionAttribute $reflectionAttribute) => $reflectionAttribute->newInstance());
+
+        $parent = $class->getParentClass();
```
rubenvanassche, line 90: Add an extra line here to give the code some breathing space

src/Support/DataClass.php
```diff
@@ -81,6 +80,21 @@ public static function create(ReflectionClass $class): self
...
+        if ($parent !== false && $class->getName() !== Data::class) {
```
rubenvanassche, line 91: You want to check on the BaseData interface here, since in v4 we'll have mutliple base Data classes like Dto and Resource.

rubenvanassche (changes requested):
Looking good, some small changes required

src/Support/DataClass.php
bastien-phi: I decided not to exclude base class provided by the package here. Easier way, not sure if we can exclude base class based on an interface it implements as the child class will also implements this interface...

bastien-phi:
@rubenvanassche This should be better

rubenvanassche:
Looking great! Thanks!

bastien-phi:
Thank you !

PR #582, opened 2023-10-09
Update creating-a-data-object.md

tsrift:
create() wants array


rubenvanassche:
You're right, thanks!

PR #588, opened 2023-10-12
feat: support multiple transformers and excluding transformers from Eloquent casting

innocenzi:
This pull request is an attempt at solving the use case described in #579. 

It implements two major changes:
- Base classes and properties now support multiple transformers
- Transformers implementing `ExcludeFromEloquentCasts` are ignored during Eloquent casting

Additionally:
- `WithTransformer` can now be applied multiple times.
- `setTransformers` has been added to `DataConfig` for convenience

Note: technically, this pull request contains a breaking change (`DataProperty::$transformer` is now `DataProperty::$transformers`).

Marking as draft because I'm not convinced about this approach.

PR #589, opened 2023-10-12
feat!: add context to transformers

innocenzi:
This pull request is an alternative to https://github.com/spatie/laravel-data/pull/588, which I personally prefer.

Along with also allowing multiple transformers for a class, it adds a context property to the `Transformer` contract, which can be an arbitrary string.

This allows to transform data differently depending on the context (`response`, `eloquent`, a custom context...).

Unfortunately, this is a breaking change if using a custom `Transformer` implementation. I couldn't find a way to implement the feature in a backwards-compatible way.

<details>
<summary>Example</summary>
&nbsp;

Usage:

```php
// In a controller
return SomeData::collection($query->paginate())->transform(context: 'hybridly');
// or
return SomeData::from($data)->transform(context: 'hybridly');
```

App-specific stuff:

```php
// Custom TypeScript Transformer that generates a proper type using `HasOptions`
final class OptionsTransformer extends EnumTransformer
{
    protected function toEnumValue(ReflectionEnumBackedCase $case): string
    {
        if ($case->getEnum()->implementsInterface(HasOptions::class)) {
            return json_encode($case->getValue()->toOption());
        }

        $value = $case->getBackingValue();

        return \is_string($value) ? "'{$value}'" : "{$value}";
    }
}

// Custom Laravel Data transformer for transforming the enum in options
final class OptionsTransformer implements Transformer
{
    public function transform(DataProperty $property, mixed $value, ?string $context = null): null|string|array
    {
        if (!$value instanceof BackedEnum) {
            return null;
        }

        if ($context === 'hybridly' && $value instanceof HasOptions) {
            return $value->toOption();
        }

        return $value->value;
    }
}

// Interface for my enums
interface HasOptions
{
    public function toOption(): array;
}

// Example implementation
enum PilotGrade: string implements HasOptions
{
    case CAPTAIN = 'captain';
    case FIRST_OFFICER = 'first_officer';
    case PILOT = 'pilot';

    public function toHumanReadableString(): string
    {
        return match ($this) {
            static::FIRST_OFFICER => 'First officer',
            static::CAPTAIN => 'Captain',
            static::PILOT => 'Pilot',
        };
    }

    public function toOption(): array
    {
        return [
            'value' => $this->value,
            'label' => $this->toHumanReadableString(),
        ];
    }
}
```

</details>

Concerns:
- `$context` as a `string` might be restrictive, this could be an array or even a dedicated class
- Calling `transform` with the same context everywhere would be repetitive, but this could be attenuated by making the `BaseData` class `Macroable`
- This is a breaking change so we have to wait until v4... :(

### DIFF
--- a/src/Attributes/WithTransformer.php
+++ b/src/Attributes/WithTransformer.php
@@ -6,7 +6,7 @@ use Attribute;
 use Spatie\LaravelData\Exceptions\CannotCreateTransformerAttribute;
 use Spatie\LaravelData\Transformers\Transformer;
 
-#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY)]
+#[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_CLASS | Attribute::TARGET_PROPERTY)]
 class WithTransformer
 {
     public array $arguments;
--- a/src/Concerns/BaseData.php
+++ b/src/Concerns/BaseData.php
@@ -102,8 +102,9 @@ trait BaseData
         bool $transformValues = true,
         WrapExecutionType $wrapExecutionType = WrapExecutionType::Disabled,
         bool $mapPropertyNames = true,
+        ?string $context = null,
     ): array {
-        return DataTransformer::create($transformValues, $wrapExecutionType, $mapPropertyNames)->transform($this);
+        return DataTransformer::create($transformValues, $wrapExecutionType, $mapPropertyNames, $context)->transform($this);
     }
 
     public function getMorphClass(): string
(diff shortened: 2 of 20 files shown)


innocenzi:
Looks like v4's `TransformationContext` might help with this feature? 👀

rubenvanassche:
Great to hear, not sure if we want an array for context but maybe that's really useable. Was also thinking about adding a TransformationOutput enum or such a thing with values : array, Model, response, json, ... not yet sure about the exact implementation.

As for this PR I'm gonna close it for now since it has a lot of conflicts and the issue seems resolved.

PR #591, opened 2023-10-13
Add DataCollectionSynth

edalzell:
Going to leave this as draft because I'd like to explain my use case and see if you folks are willing to accept this. If so, I will add tests.

In my project I have a custom DataCollection called `StationCollection` and because I'm using Livewire (v3), I had it `implement Wireable`. However, the `WireableSynth` in Livewire expects properties:

```php
    function set(&$target, $key, $value) {
        $target->{$key} = $value;
    }
```

Which throws an error when used.

To work around this I created a `DataCollectionSynth` that handles things properly, and then made a custom synth that inherits from `DataCollectionSynth` to deal with my custom stuff (mostly type hinting).

Note that to use this, one would have to add this to your service provider, `Livewire::propertySynthesizer(DataCollectionSynth::class);`

Does that sound reasonable?

### DIFF
--- a/src/Synths/DataCollectionSynth.php
+++ b/src/Synths/DataCollectionSynth.php
@@ -28,9 +28,19 @@ class DataCollectionSynth extends Synth
         ];
     }
 
-    public function hydrate($value, $meta)
+    /**
+     * @param  array<mixed>  $value
+     * @param  array<string, class-string>  $meta
+     * @param  mixed  $hydrateChild
+     * @return \Spatie\LaravelData\DataCollection
+     */
+    public function hydrate($value, $meta, $hydrateChild)
     {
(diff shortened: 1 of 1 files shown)


rubenvanassche:
Hi @edalzell,

Thanks for the PR, I first need to get started with LiveWire 3 and read myself into the hydration and synthesizers part. When that's done I'll again take a look at this PR!

edalzell:
Just pushed a fix, we need to dehydrate recursively (thanks [Josh](https://github.com/joshhanley)!)

rubenvanassche:
Still on my radar, now I have some time since v4 is our gonna take a look at this next week or the week after that.

rubenvanassche:
Thanks!

PR #597, opened 2023-11-01
Add generic type for WithData trait

konovalexander:
It becomes possible to add PHPDoc when using the "WithData" trait:
```
class CommentRequest extends FormRequest {

    /** @use WithData<\App\Data\CommentData> */
    use WithData;

    protected string $dataClass = CommentData::class;

    //Some code
}
```
Then type hinting works correctly after calling the "getData" method:
```
public function store(CommentRequest $request): JsonResponse {
    
    $commentData = $request->getData();
    //$commentData is now of type CommentData
    
    //Some code
}
```

### DIFF
--- a/src/WithData.php
+++ b/src/WithData.php
@@ -3,12 +3,17 @@
 namespace Spatie\LaravelData;
 
 use Spatie\LaravelData\Contracts\BaseData;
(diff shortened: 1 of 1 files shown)


rubenvanassche:
Fancy 😍 , thanks!

Issue #583, opened 2023-10-10
Non-self explanatory Optional property initialization


raveren:
**✏️ Describe the bug**

This works:
```php
class SongData extends Data
{
    public string $title;
    public array|Optional $artist;

    public function __construct() {}
}

class DataTest extends TestCase
{
    public function test_case: void
    {
        $s        = SongData::from([]);
        $s->title = 'ads';

        dd($s->toArray());
    }
}
```

This doesn't:

```
Error: Typed property Tests\App\SongData::$artist must not be accessed before initialization
/var/www/vendor/spatie/laravel-data/src/Transformers/DataTransformer.php:79
/var/www/vendor/spatie/laravel-data/src/Transformers/DataTransformer.php:48
/var/www/vendor/spatie/laravel-data/src/Concerns/BaseData.php:106
/var/www/vendor/spatie/laravel-data/src/Concerns/TransformableData.php:16
```

```php
class SongData extends Data
{
    public string $title;
    public array|Optional $artist;

    public function __construct() {}
}

class DataTest extends TestCase
{
    public function test_case: void
    {
        $s        = SongData::from(); // <-- removed empty array
        $s->title = 'ads';

        dd($s->toArray());
    }
}
```

And most importantly this, natural use, which would allow usage of LaravelData without any documentation doesn't work either:

```php
class SongData extends Data
{
    public string $title;
    public array|Optional $artist;

    public function __construct() {}
}

class DataTest extends TestCase
{
    public function test_case: void
    {
        $s        = new SongData(); // <-- 
        $s->title = 'ads';

        dd($s->toArray());
    }
}
```

**✅ Expected behavior**

Logically, the class initialization should not alter this behaviour, the initialization should yield equivalent results in the end.

Especially the `DataClass::from()` vs `DataClass::from([])` is a very niche solution for this, feels like a work-around.

**🖥️ Versions**

Laravel: latest
Laravel Data: latest
PHP: latest

rubenvanassche:
Yeah weird behavior, fixed it in the version released today.

rubenvanassche:
Reverted due to the implementation being more difficult than expected: #590

ronssij:
@raveren from what I have understand if the field is optional, sometimes it could be null and it is a realtionship. you can do the `Data|Lazy|Optional`

```
class SongData extends Data
{
    public string $title;
    public ArtistData|Lazy|Optional $artist;

    public function __construct() {}
}
```

raveren:
[...]

Issue #590, opened 2023-10-13
[3.9.1] backward compatibility is broken


alexrififi:
**✏️ Describe the bug**

I have LocationData with

```php
    public function __construct(
        public int|null|Optional $parent_id,
    ) {
    }
```

**Optional** is present for partial API update as [in the documentation](https://spatie.be/docs/laravel-data/v3/as-a-data-transfer-object/optional-properties)

In version `3.9.0`, if `parent_id` is `null` in the Model, then in the data object `parent_id` is also equal to `null`.

In version `3.9.1`, if `parent_id` is `null` in the Model, then in the data object `parent_id` is absent

**✅ Expected behavior**

If `parent_id` is `null` in the Model, then in the data object `parent_id` also to be equal to `null`. Otherwise, it is not clear how to use `Optional` in this case (when there may be a partial API update, but null in the model also matters)

**🙏 The reason**

It looks like this commit https://github.com/spatie/laravel-data/commit/3b63875338cd6ef29fe362421a4e2186338f5b2d

**🖥️ Versions**

Laravel: 10.28.0
Laravel Data: 3.9.1
PHP: 8.2

Ferikl:
Same issue!

rubenvanassche:
Fixed in 3.9.2

PR #602, opened 2023-11-08
Fix issue where non set optional values were transformed

rust17:
fix: #583 and be compatible with backward which mentioned in #590

### DIFF
--- a/src/Transformers/DataTransformer.php
+++ b/src/Transformers/DataTransformer.php
@@ -76,6 +76,10 @@ class DataTransformer
 
             $name = $property->name;
 
+            if ($property->type->isOptional && ! array_key_exists($name, get_object_vars($data))) {
+                continue;
+            }
+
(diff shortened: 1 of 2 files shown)


rust17:
@rubenvanassche 😃  I think it's helpful. If you think this change useful too, please consider merging this.

rubenvanassche:
Good solution, never would have though about that one. Hopefully it will not break performance too hard. Thanks

xHeaven:
@rubenvanassche Probably just a nitpick, also didn't test it at all so I'm just talking out of my head right now: do we really want to do `get_object_vars()` inside the `foreach` loop? Can't we do `$objectVars = get_object_vars($data);` or something like this _outside_ the loop, and referencing that variable inside the loop? I think it'd help a little if you have massive datasets if we didn't query all the object vars in every iteration from the exact same dataset.

PR #630, opened 2023-12-31
Edit the config file to specific the namespace and the suffix

alihesahm:
I think adding this configuration will help the developer to know that they can config the namespace and the suffix of the generated Data class


rubenvanassche:
Yeah we didn't add this in the past to keep the config file clear, there is some documentation about this in the docs

PR #642, opened 2024-01-21
Update working-with-dates.md

ricardov03:
Adding the PHP documentation with the different DateTimeInterface predefined constanst.


rubenvanassche:
I think this is a quick google away, I wanna keep the documentation as clear as it is.

PR #643, opened 2024-01-21
[add] Arr::flatter wrap methods params

pepperfm:
I always wanted to pass parameters as array to `only()` and `except()` method to make code more light and slim on break lines

Here is a solution: now methods accept string and array the same way


rubenvanassche:
Why not do `include(...$myArray)`?

PR #648, opened 2024-01-24
Added `push()` method

Plytas:
I was looking for a fluent way to add items to collection but wasn't able to find one.
Current alternatives:
```php
SimpleData::collection()->toCollection()->push('foo'); //Technically works if underlying collection is `\Illuminate\Support\Collection`, but cannot be properly type hinted because `$items` is `\Illuminate\Support\Enumerable`.
```
```php
$collection = SimpleData::collection();
$collection[] = 'foo'; //This works because `DataCollection` implements `ArrayAccess`
```

This proposed method is the same as `\Illuminate\Support\Collection::push()`.

### DIFF
--- a/src/DataCollection.php
+++ b/src/DataCollection.php
@@ -78,6 +78,20 @@ class DataCollection implements DataCollectable, ArrayAccess
     }
 
     /**
+     * @param TValue ...$values
+     *
+     * @return static
+     */
+    public function push(...$values): static
+    {
+        foreach ($values as $value) {
+            $this[] = $value;
+        }
+
+        return $this;
+    }
+
+    /**
      * @param TKey $offset
      *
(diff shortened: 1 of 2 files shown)


rubenvanassche:
In v4 these methods are deprecated since we want to move away from DataCollections to regular Collections, we're not going to add any extra methods in such case just use a Collection from now on.

Plytas:
Wasn't aware of upcoming v4 changes and I see it has been tagged recently. Thank you!

rubenvanassche:
Yeah has been in the making for a year, glad I've could finally release it 😄 

I hope to completely remove DataCollection in v5 but that requires some extra thinking about how to handle a few scenarios, that's something for the next weeks.

Issue #660, opened 2024-02-11
Incorrect CastPropertiesDataPipe logic


alexrififi:
**✏️ Describe the bug**
In the case where the cast returns an empty array, it is not returned

**↪️ To Reproduce**
```php
use Illuminate\Support\Collection;
use Spatie\LaravelData\Attributes\WithCast;
use Spatie\LaravelData\Casts\Cast;
use Spatie\LaravelData\Data;
use Spatie\LaravelData\Support\Creation\CreationContext;
use Spatie\LaravelData\Support\DataProperty;

it('cannot cast empty collection', function () {
    class KeysCast implements Cast
    {
        public function cast(
            DataProperty $property,
            mixed $value,
            array $properties,
            CreationContext $context
        ): array {
            return $value instanceof Collection ? $value->keys()->all() : $value;
        }
    }

    class SomeData extends Data
    {
        public function __construct(
            #[WithCast(KeysCast::class)]
            public array $items,
        ) {
        }
    }

    // normal behavior
    $data = SomeData::from(['items' => new Collection(['one' => 'value 1', 'two' => 'value 2'])]);
    expect($data)->toBeInstanceOf(SomeData::class);
    expect($data->items)->toEqual(['one', 'two']);

    // TypeError: SomeData::__construct(): Argument #1 ($items) must be of type array, Illuminate\Support\Collection given
    $data = SomeData::from(['items' => new Collection()]);
    expect($data)->toBeInstanceOf(SomeData::class);
    expect($data->items)->toEqual([]);
});
```

**✅ Expected behavior**
I expect that the data's property `items` will equal an empty array

**🖥️ Versions**

Laravel: 10.43.0
Laravel Data: 4.0.1
PHP: 8.3.1

---

## The problem is here

https://github.com/spatie/laravel-data/blob/67d279485fccb78ada54405bd347bf602697c967/src/DataPipes/CastPropertiesDataPipe.php#L55-L59

When `$casted` is equal to an empty array, it is not returned

## Note

I also assume that if the cast returns a value that is [**interpreted as false**](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting), in some cases it may break the application's logic.

Skullbock:
Adding another case here: If i have a custom cast that may return `null`, which for that casts is a valid value, this same data pipeline will not return null and keep the original value, for the same reasoning

rubenvanassche:
Should be fixed with the next release!

PR #662, opened 2024-02-11
[v4] Fix CastPropertiesDataPipe logic

alexrififi:
This PR attempts to fix https://github.com/spatie/laravel-data/issues/660


rubenvanassche:
Thanks for the starting point, that was really handy! Eventually I've made my own implementation should be released later today!

PR #665, opened 2024-02-12
Prevent out of memory with multiple requests in a test

### LINKED ISSUE #656: [V4] Out of memory when resolving Data object from container during feature tests
**✏️ Describe the bug**

I have a feature test that makes requests to a controller that is resolving a Data object from the container. The second request causes an out of memory error. If I only make one request in the test it works fine.

**↪️ To Reproduce**

```php
class MyController extends Controller
{
    // method injection causes error
    public function __invoke(SongData $data)
    //public function __invoke()
    {
        // this causes error
        //$data = app(SongData::class);

        // this does not cause error
        //$data = SongData::from(request());

        return response('ok');
    }
}
```

```php
class SongData extends Data
{
    public function __construct(
      public string $title,
    ) {}
}
```

```php
class MyTest extends TestCase
{
    public function test_example(): void
    {
        $this->post('/test', ['title' => 'test'])->assertOk();
        $this->post('/test', ['title' => 'test'])->assertOk();
    }
}
```

```
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /testapp/vendor/laravel/framework/src/Illuminate/Container/Container.php on line 272

Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /testapp/vendor/laravel/framework/src/Illuminate/Container/Container.php on line 272
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes) in /testapp/vendor/symfony/error-handler/Error/FatalError.php on line 1

Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes) in /testapp/vendor/symfony/error-handler/Error/FatalError.php on line 1
```

**✅ Expected behavior**

I expect the test to pass and to not run out of memory.

**🖥️ Versions**

Laravel: 10.43.0
Laravel Data: 4.0.1
PHP: 8.3.2

ejunker:
This fixes #656. I added a test and got it to pass by restoring some code from the v3 branch.

Here is the [previous commit](https://github.com/spatie/laravel-data/commit/46e2d19c8bfebe1bca4fc5c7149ab21bac9461fb#diff-66963ca1bf7d4336f8d577f1113c285b8be2aae97a64d22f983c75acf12c1285) when the check was removed.

I do not use Laravel Octane or know enough about it to know if this fix would be compatible with Octane.

### DIFF
--- a/src/LaravelDataServiceProvider.php
+++ b/src/LaravelDataServiceProvider.php
@@ -41,6 +41,10 @@ class LaravelDataServiceProvider extends PackageServiceProvider
         );
 
         $this->app->beforeResolving(BaseData::class, function ($class, $parameters, $app) {
+            if ($app->has($class)) {
+                return;
+            }
+            
             $app->bind(
                 $class,
                 fn ($container) => $class::from($container['request'])
--- a/tests/RequestTest.php
+++ b/tests/RequestTest.php
@@ -21,7 +21,7 @@ beforeEach(function () {
         ValidationException::class,
     ]);
 
-    Route::post('/example-route', function (Sim $data) {
+    Route::post('/example-route', function (SimpleData $data) {
         return ['given' => $data->string];
     });
 });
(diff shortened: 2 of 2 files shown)


jameshulse:
Nice work finding a fix for this - we are running into the same issue which made the upgrade to v4 impossible.

If that conditional is incompatible with Octane, it could be possible to check if Octane is running with something similar to this: https://github.com/laravel/octane/issues/508#issuecomment-1096937174

PR #669, opened 2024-02-19
laravel-data & Livewire

rubenvanassche:
A set of changes to make laravel-data and Livewire work better together, very wip


ju5t:
Would you consider adding validation too? If possible, it would also be awesome if we can validate nested data. Something along the lines of:

```php
public AlbumData $album;

public function save()
{
    $this->album->validate();
    //
}
```

Our plan is to use data objects instead of Livewire's form objects so we can reuse it in our upcoming API.

ju5t:
@rubenvanassche how would data validation work when this is implemented? Will there be no validation or are you validating data before the data object gets updated?

rubenvanassche:
No validation, it is at the moment technically not possible to provide correct validation objects. You'll need to implement it yourself.

PR #671, opened 2024-02-19
Backport cache fix to v3

jameshulse:
I noticed that the latest V4 version has a fix to caching - is it possible to back port this to V3 while we are still on this version? We can't easily upgrade to V4 yet, but I have noticed that the laravel-data caching isn't particularly performant in production and maybe this will help.

I also noticed there was a separate caching fix commit, but I'm not sure if it is applicable to V3 as it had a lot more change.


rubenvanassche:
Thanks!

PR #675, opened 2024-03-01
Update wrapping.md to clarify collections and paginated collections case

santigarcor:
The current docs don't clarify we need to specifically collect data into `DataCollection` or `PaginatedDataCollection` in order to set a custom wrapper


rubenvanassche:
Thanks!

PR #682, opened 2024-03-05
fix: support arrays when transforming paginator data

innocenzi:
This pull request fixes an error that I am encountering when serializing a paginated data collection: https://flareapp.io/share/LPdKl3dm

I haven't got to the bottom of it by lack of time, but the gist is that the following does not work:

```php
public function index()
{
    $chirps = Chirp::query()
        ->forHomePage()
        ->paginate();

    return hybridly('chirps.index', [
        'chirps' => ChirpData::collect($chirps, PaginatedDataCollection::class),
    ]);
}
```

While this does work:

```php
public function index()
{
    $chirps = Chirp::query()
        ->forHomePage()
        ->paginate();

    return hybridly('chirps.index', [
        'chirps' => ChirpData::collect($chirps, PaginatedDataCollection::class)->transform(),
    ]);
}
```

Notice how the collection is manually transformed there. 

I'm not sure yet why this happens, but I figured due to the condition in `transformationClosure` that the `$data` argument could be something else than a `BaseData` anyway, which would fix the issue.

### DIFF
--- a/src/Resolvers/TransformedDataCollectableResolver.php
+++ b/src/Resolvers/TransformedDataCollectableResolver.php
@@ -82,7 +82,7 @@ class TransformedDataCollectableResolver
         Wrap $wrap,
         TransformationContext $nestedContext,
     ): array {
-        $paginator = $paginator->through(fn (BaseData $data) => $this->transformationClosure($nestedContext)($data));
+        $paginator = $paginator->through(fn (array|BaseData $data) => $this->transformationClosure($nestedContext)($data));
 
         if ($nestedContext->transformValues === false) {
             return $paginator->all();
@@ -105,7 +105,7 @@ class TransformedDataCollectableResolver
     protected function transformationClosure(
         TransformationContext $nestedContext,
     ): Closure {
-        return function (BaseData $data) use ($nestedContext) {
+        return function (array|BaseData $data) use ($nestedContext) {
             if (! $data instanceof TransformableData || ! $nestedContext->transformValues) {
                 return $data;
             }


rubenvanassche:
Can you tell me what the object looks like (or @LukeAbell), tried replicating this:

```php
it('bla', function (){
    Route::get('/', function (){
        FakeModel::factory()->count(10)->create();

        $p = FakeModelData::collect(FakeModel::paginate(), PaginatedDataCollection::class);

        return [
            'data' => $p,
        ];
    });

    getJson('/')->dump()->assertOk();
});
```

But it doesn't fail, maybe nested objects/collections are causing the issue?

I get that this PR is the "easy" solution, but I would expect Data objects in that closure so I wanne first see if we can fix the issue upstream.

innocenzi:
Hey Ruben, here's the object in question: https://github.com/hybridly/demo/blob/feat/upgrade-hybridly/app/Data/ChirpData.php

The reproduction is unfortunately not straightforward, but if you're feeling like it:

- Clone `hybridly/hybridly` and install deps
- Build `hybridly/hybridly` using `npm run build`
- Clone the `feat/hybridly-upgrade` branch of `hybridly/demo`
- In `composer.json` replace the path in the `repositories` object with the path to `packages/laravel` of your own local `hybridly/hybridly` clone
- Run the demo project
- Log in (there is a quick login link)
- The `/` page will trigger the error (`app/Http/Controllers/ChirpController.php`)

rubenvanassche:
Hi @innocenzi,

Can't really say what's going wrong with your projects, I do not have the time atm to check it all out.

We had similar issues in an old project we've updated:

- Sometimes an array was typed to have Data objects in it, but there were arrays in it causing issues
- Sometimes the PaginatedDataCollection through method returned arrays, again the same issue this is not allowed

We've only added more logic to these functions in the last days, when we would allow arrays here other functionalities would not work. That's why we're going to keep it on the BaseData interface type. So that we have one clear path in out code what's happening.

innocenzi:
Right, I understand. I'll investigate and if I find out that the issue indeed lies in Data, I'll open another PR. Thank you for your time ❤️

rubenvanassche:
Good luck with it! These kinds of issues are sadly enough hard to debug 😞

innocenzi:
Note to self—after some more debugging today, I got the following error: https://flareapp.io/share/lm2vKxjm

This time, it's the VarDumper that throws the exception (and not my app or Hybridly). There was an unrelated exception in my code, and when trying to serialize it, the VarDumper threw this exception related to serializing Data.

Still looking into it :)

innocenzi:
OKAY I finally got it. Basically, transforming a `PaginatedDataCollection` mutates the internally-wrapped `LengthAwarePaginator` because of the `through` method.

Now, my code has a `$properties` array (think Inertia properties) that is transformed twice to resolve some stuff inside of it.

But during the transformation, the `LengthAwarePaginator` is mutated from Data objects to arrays. So the second transformation throws because it doesn't accept arrays, even though I wasn't using the result of the first transformation but the initial variable 🫠

PR #693, opened 2024-03-12
Handle `null` for `fieldContext` within `resolvePotentialPartialArray`

faustbrian:
When `resolvePropertyValue` calls `resolvePotentialPartialArray` there is an exception thrown if `$fieldContext` is null. This is possible because the parameters of `resolvePropertyValue` contain `?TransformationContext $fieldContext` which means it can be `null` but wasn't allowed to be null within `resolvePotentialPartialArray`.

**This fixes that and was tested by making our own test suite pass again and errors going away when using our application.**


rubenvanassche:
Thanks!

PR #694, opened 2024-03-12
Support duration in cache configuration

sebastiandedeyne:
The default duration is `null` which will cache forever, which is the current behavior.


rubenvanassche:
Thanks!

PR #696, opened 2024-03-12
fix: avoid mutation during `DataCollectable` transformation

innocenzi:
This pull request is a follow-up of https://github.com/spatie/laravel-data/pull/682.

The issue was that paginators inside `DataCollection` classes were mutated upon transformation, resulting in an exception when trying to transform them again.

### DIFF
--- a/src/Resolvers/TransformedDataCollectableResolver.php
+++ b/src/Resolvers/TransformedDataCollectableResolver.php
@@ -82,7 +82,7 @@ class TransformedDataCollectableResolver
         Wrap $wrap,
         TransformationContext $nestedContext,
     ): array {
-        $paginator = $paginator->through(fn (BaseData $data) => $this->transformationClosure($nestedContext)($data));
+        $items = array_map(fn (BaseData $data) => $this->transformationClosure($nestedContext)($data), $paginator->items());
 
         if ($nestedContext->transformValues === false) {
             return $paginator->all();
@@ -93,7 +93,7 @@ class TransformedDataCollectableResolver
         $wrapKey = $wrap->getKey() ?? 'data';
 
         return [
-            $wrapKey => $paginated['data'],
+            $wrapKey => $items,
             'links' => $paginated['links'] ?? [],
             'meta' => Arr::except($paginated, [
                 'data',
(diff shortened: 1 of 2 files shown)


src/Resolvers/TransformedDataCollectableResolver.php
```diff
@@ -82,7 +82,7 @@ protected function transformPaginator(
         Wrap $wrap,
         TransformationContext $nestedContext,
     ): array {
-        $paginator = $paginator->through(fn (BaseData $data) => $this->transformationClosure($nestedContext)($data));
+        $items = array_map(fn (BaseData $data) => $this->transformationClosure($nestedContext)($data), $paginator->items());
 
         if ($nestedContext->transformValues === false) {
             return $paginator->all();
```
rubenvanassche, line 88: I think somehow, we also should be able to somehow run the items here through the closure? If you pull in the latest changes of this file you'll see that the `transformationClosure` is a bit more complicated than it was. Maybe we can use through here? Since we're not transforming values?

rubenvanassche (changes requested):

src/Resolvers/TransformedDataCollectableResolver.php
innocenzi: I'm sorry, I'm not sure what you mean?

rubenvanassche: The items in:

```php
 return $paginator->all();
```

Will now not run through the `transformationClosure`, which may cause problems since the new version of `transformationClosure` also performs operations on the data objects even if we're not transforming values.

innocenzi: Okay, could that be done in another PR? If you'd rather have it in this PR, I'm not sure what the behavior should be. Remove the guard clause in `transformPaginator`? Or adapt the early return to return the mapped items?

rubenvanassche: I think calling this should be enough:

```php
 return $paginator->through(fn (BaseData $data) => $this->transformationClosure($nestedContext)($data))->all();
```

rubenvanassche:
Thanks

PR #697, opened 2024-03-13
docs: fix throw Uncastable in quickstart Local Cast example;

### LINKED ISSUE #685: throw Uncastable results in "Cannot throw objects that do not implement Throwable" from a custom Cast
**✏️ Describe the bug**
Per [docs](https://spatie.be/docs/laravel-data/v4/getting-started/quickstart#content-local-casts) it says you need to `throw Uncastable::create();` to tell the package to try other casts, but the real result is that for PHP 8.2 it throws an `\Error` with the message `Cannot throw objects that do not implement Throwable`. Not to mention every static analysis tells me it's wrong. 

**↪️ To Reproduce**
Create a custom cast -> use it so that it definitely throws Uncastable. 

```php
use Spatie\LaravelData\Attributes\WithCast;
use Spatie\LaravelData\Casts\Cast;
use Spatie\LaravelData\Casts\Uncastable;
use Spatie\LaravelData\Data;
use Spatie\LaravelData\Support\Creation\CreationContext;

class StringCast implements Cast
{
    public function cast(DataProperty $property, mixed $value, array $properties, CreationContext $context): string
    {
        throw Uncastable::create();
        return 'anything';
    }
}

class PostData extends Data
{
    public function __construct(
        #[WithCast(StringCast::class)]
        public string $myString,
    ) {
    }
}

// here it throws the \Error
$postData = PostData::from([
    'myString' => 'hello',
]);
```

**✅ Expected behavior**
The Uncastable class to implement \Thorwable.

By the way how to achieve this "other casts" thing? I've tried to add another `#[WithCast(...)]` but it results in an \Error with message `Attribute "Spatie\LaravelData\Attributes\WithCast" must not be repeated`. 

**🖥️ Versions**

Laravel: v10.22.0
Laravel Data: 4.1.0
PHP: 8.2.4

olexoliinyk0:
closes #685;

### DIFF
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -310,7 +310,7 @@ use Str;
 
 class ImageCast implements Cast
 {
-        public function cast(DataProperty $property, mixed $value, array $properties, CreationContext $context): Image
+        public function cast(DataProperty $property, mixed $value, array $properties, CreationContext $context): Image|Uncastable
     {
         // Scenario A
         if ($value instanceof UploadedFile) {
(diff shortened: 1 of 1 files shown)


rubenvanassche:
Thanks!

PR #702, opened 2024-03-13
fix: not ignore magical creation for nested DTOs;

### LINKED ISSUE #701: Magical creation is not called on the nested DTOs
**✏️ Describe the bug**
When using [magical creation](https://spatie.be/docs/laravel-data/v4/as-a-data-transfer-object/creating-a-data-object#magical-creation) "from..." it won't be called on the nested DTOs, but should have. 

**↪️ To Reproduce**
Create some base (or simple) DTO, that has magical creation, for example with an array $payload, and add some logic there that should fail on incorrect payload.

```php
class SomeDto extends Data
{
    public string $name;

    public Optional|SomeDto $nestedItem;

    /** @var array<int, SomeDto> */
    public array $nestedArrayOfItems;

    public static function fromSome(array $payload, CreationContext $creationContext): static
    {
        $factory = static::factory($creationContext);
        $factory->ignoreMagicalMethod('fromSome');
        $dto = $factory->from($payload);

        if (!isset($dto->name)) {
            throw new \TypeError('"name" is not set'); // <-- should throw this on every SomeDto creation, even for nested ones
        }

        return $dto;
    }
}

// should check if "name" set on every level for every SomeDto, but it does not
$someDto = SomeDto::from([
    'name' => 'some', // for this level it will check if "name" is set
    'nestedItem' => [], // for this level it will skip "fromSome" at all, because of the same CreationContext with `ignoreMagicalMethod`
    'nestedArrayOfItems' => [
        [],  // same here
    ],
]);

dd($someDto->toArray());
// [
//   "name" => "some"
//   "nestedItem" => []
//   "nestedArrayOfItems" => [
//     0 => []
//   ]
// ]
```

**✅ Expected behavior**
The `CastPropertiesDataPipe` should not use parent's `$creationContext`, where we already added our "fromSome" to `ignoredMagicalMethods`, but should check every nested DTOs with a new CreationContext, that will not ignore our magical method.

**🖥️ Versions**

Laravel: v10.22.0
Laravel Data: 4.1.0
PHP: 8.2.4

olexoliinyk0:
fixes #701;

I might not see the reasons for it to use the same CreationContext, but it seems to me that there's no one.

There's also some CreationContext next/previous jumping in the `Spatie\LaravelData\Resolvers\DataCollectableFromSomethingResolver::itemsToDataClosure` but I can't tell if it is connected to the issue.

### DIFF
--- a/src/DataPipes/CastPropertiesDataPipe.php
+++ b/src/DataPipes/CastPropertiesDataPipe.php
@@ -85,11 +85,13 @@ class CastPropertiesDataPipe implements DataPipe
             $property->type->kind->isDataObject()
             || $property->type->kind->isDataCollectable()
         ) {
-            $context = $creationContext->next($property->type->dataClass, $property->name);
+            $propertyCreationContextFactory = CreationContextFactory::createFromConfig($property->type->dataClass);
+            $propertyCreationContext = $propertyCreationContextFactory->get();
+            $propertyCreationContext->next($property->type->dataClass, $property->name);
 
             return $property->type->kind->isDataObject()
-                ? $context->from($value)
-                : $context->collect($value, $property->type->iterableClass);
+                ? $propertyCreationContext->from($value)
+                : $propertyCreationContext->collect($value, $property->type->iterableClass);
         }
 
         if (
--- a/src/DataPipes/CastPropertiesDataPipe.php
+++ b/src/DataPipes/CastPropertiesDataPipe.php
@@ -9,6 +9,7 @@ use Spatie\LaravelData\Enums\DataTypeKind;
 use Spatie\LaravelData\Lazy;
 use Spatie\LaravelData\Optional;
 use Spatie\LaravelData\Support\Creation\CreationContext;
+use Spatie\LaravelData\Support\Creation\CreationContextFactory;
 use Spatie\LaravelData\Support\DataClass;
 use Spatie\LaravelData\Support\DataConfig;
 use Spatie\LaravelData\Support\DataProperty;


olexoliinyk0:
Came to think about it again, and it breaks the `$craetionContext->currentPath` consistency. 

Can't just yet tell how to fix that. 

#### Side note, possibly a bug:
- in the **main** branch. 
- if you add [here](https://github.com/spatie/laravel-data/blob/09b38e5388a90effe4d254f3e805bda7ee20bd4b/src/DataPipes/CastPropertiesDataPipe.php#L186) `dump($context->currentPath);` after the line.
- and try that example from #701 but with this data 

```php
$someDto = SomeDto::from([
    'name' => 'some',
    'nestedItem' => [
        'name' => 'nestedItem-1',
    ],
    'nestedArrayOfItems' => [
        [
            'name' => 'nestedArrayOfItems-1',
            'nestedItem' => [
                'name' => 'nestedArrayOfItems-1--nestedItem-1',
            ],
        ],
    ],
]);
```
- it outputs 
   - `['nestedItem']`
   - `['nestedItem', 'nestedArrayOfItems']`
   - `['nestedItem', 'nestedArrayOfItems', 0, 'nestedItem']`
- but should (IMO) output 
   - `['nestedItem']`
   - `['nestedArrayOfItems']`
   - `['nestedArrayOfItems', 0, 'nestedItem']`

So it does not go back (`$context->previous();`) after going into the first `nestedItem` (_nestedItem-1_).

rubenvanassche:
As said in the issue, there's a reason why contexts are recycled. We're going to keep that.

PR #707, opened 2024-03-15
Cast value to string for createFromFormat

### LINKED ISSUE #708: Can no longer use parse integer timestamp into Carbon object with Laravel 11
**✏️ Describe the bug**
As of Carbon v3, they have introduced strict typing throughout the project, this has caused the `createFromFromat` function used by the DateTimeInterfaceCast class in this library to throw an error when it is provided an integer.

```
Carbon\CarbonImmutable::rawCreateFromFormat(): Argument #2 ($time) must be of type string, int given, called in /home/vagrant/atlas/vendor/nesbot/carbon/src/Carbon/Traits/Creator.php on line 647
```

I have an issue open with Carbon about this as well: https://github.com/briannesbitt/Carbon/issues/2967

As a workaround I wrote my owner caster specific to CarbonInterface objects that uses the `parse` function to interpret the timestamps instead and this seems to work fine.

**↪️ To Reproduce**

```php

it('cannot parse a unix timestamp into a carbon object', function () {
    class DateDtoTest extends Data
    {
        public function __construct(
            public CarbonImmutable $date,
        ) {
        }
    }

    dd(DateDtoTest::from(['date' => 1710541965]));
});
```

Additionally you will need to add `U` as an accepted date time format to your configuration file.

**✅ Expected behavior**
Allow the parsing of an integer timestamp using the built in DateTimeInterfaceCast class.

**🖥️ Versions**

Laravel: 11.0.5
Laravel Data: 4.4.0
PHP: 8.3

kylekatarnls:
Cast `$value` to `string` explictly because `DateTime::createFromFormat()` expects `string` since PHP 8

Fix #708

### DIFF
--- a/src/Casts/DateTimeInterfaceCast.php
+++ b/src/Casts/DateTimeInterfaceCast.php
@@ -42,7 +42,7 @@ class DateTimeInterfaceCast implements Cast, IterableItemCast
         $datetime = $formats
             ->map(fn (string $format) => rescue(fn () => $type::createFromFormat(
                 $format,
-                $value,
+                (string) $value,
                 isset($this->timeZone) ? new DateTimeZone($this->timeZone) : null
             ), report: false))
             ->first(fn ($value) => (bool) $value);


kylekatarnls:
Hello 👋

I know it's not `spatie/laravel-data` responsibility to explicit this call since since you don't use `declare(strict_types=1)` in this file. But I have some tricky constraints with Carbon upgrade from 2 to 3. And it fails trying to do:

```php
Carbon::createFromFormat('U', $timestamp)
```

With `$timestamp` is `int`.

Elnadrion:
Is this PR going to be merged any soon? My team has exact same problem after bumping laravel-data, carbon and laravel to the next major versions

kylekatarnls:
@rubenvanassche 🙏 I'm sorry this strict type change affect your library. I will evaluate if I can do smarter validation such as accepting and casting int if the format is a single letter that expect a numeric value.

But meanwhile a cast would be the safer fix.

Elnadrion:
@kylekatarnls as a temp solution for the issue we created custom caster:
```php
<?php

namespace App\Casters;

use DateTimeInterface;
use Spatie\LaravelData\Casts\DateTimeInterfaceCast;
use Spatie\LaravelData\Casts\Uncastable;
use Spatie\LaravelData\Support\Creation\CreationContext;

class CustomDateTimeInterfaceCast extends DateTimeInterfaceCast
{
    public function cast(DataProperty $property, mixed $value, array $properties, CreationContext $context): DateTimeInterface|Uncastable
    {
        return parent::cast($property, (string)$value, $properties, $context);
    }

    public function castIterableItem(DataProperty $property, mixed $value, array $properties, CreationContext $context): DateTimeInterface|Uncastable
    {
        return parent::castIterableItem($property, (string)$value, $properties, $context);
    }
}

```

rubenvanassche:
Thanks!

PR #712, opened 2024-03-18
Fix pagination documentation

Tofandel:
Without passing `PaginatedDataCollection::class` into collect since v4, the data would not be presented as in the documentation

### DIFF
--- a/docs/as-a-resource/from-data-to-resource.md
+++ b/docs/as-a-resource/from-data-to-resource.md
@@ -52,7 +52,7 @@ Will return a collection automatically transformed to JSON:
 It is also possible to provide a paginator:
 
 ```php
-SongData::collect(Song::paginate());
+SongData::collect(Song::paginate(), PaginatedDataCollection::class);
 ```
 
 The data object is smart enough to create a paginated response from this with links to the next, previous, last, ... pages:


rubenvanassche:
These docs are correct, you can indeed provide a PaginatedDataCollection. But in the end a Paginator will be returned with data items in it.

Tofandel:
See the data given in the example which shows pagination data under `meta`
https://github.com/spatie/laravel-data/pull/712/files#diff-29ec7e778018601e2470d394fd571f7b81141b19f08d9c4264208ba47d2c2b6eR72-R78

Unless you give a PaginatedDataCollection::class as a second argument you will not get this data in v4 (but in v3 you did)

So unless this is bug the documentation is not correct

rubenvanassche:
This is more a Laravel issue, since Laravel will represent a JSON version of a paginator differently when returned raw or as a Resource object. The `PaginatedDataCollection` version is more like a Laravel Resource. Can you update the docs accordingly to this? So keep the part about the paginator as is and add an extra part below that explains why to use a PaginatedDataCollection.

Tofandel:
How is it a laravel issue?

https://github.com/spatie/laravel-data/blob/09b38e5388a90effe4d254f3e805bda7ee20bd4b/src/Resolvers/TransformedDataCollectableResolver.php#L105-L108

This is the code that takes care of outputing the meta of the paginator, it doesn't run if you don't put 'PaginatedDataCollection::class` as the second argument

rubenvanassche:
Because that code is specifically written to be in line with Laravel Resources, try outputting a regular paginator as JSON in Laravel without using resources. You'll see the format is exactly the same as without the `PaginatedDataCollection::class` argument.

As said earlier, Laravel has two formats to represent the same paginator a bit weird that's why PaginatedDataCollection exists.

williamjulianvicary:
Jumping in here - I've ran into a non-documented regression during an upgrade from v3 > v4 and this is exactly as described.

The documentation very clearly states how it should work (and did used to work) and this is no longer the case since the upgrade.

Thanks for the tip on the second param - I'll run with that.

carlosmascarenhas:
I know this is closed, but I ran into the exact same issue and want to leave a note for anyone who lands here.

The documentation shows that passing a paginated collection to `collect()` is enough to get a structured JSON response with `meta` and `links`. That was true in v3, but in v4 it no longer works — `collect($paginator)` returns the raw Laravel paginator, which serializes to a flat JSON format.

To get the documented behaviour in v4, the second argument is required:

```php
// v4 — flat output (undocumented breaking change)
InstallmentData::collect(BillInstallment::paginate());

// v4 — structured output with meta and links (matches the docs)
InstallmentData::collect(BillInstallment::paginate(), PaginatedDataCollection::class);
```

Would it be worth updating the v4 documentation to reflect this? The current wording still implies the smart detection works out of the box, which trips up anyone migrating from v3.

PR #714, opened 2024-03-19
Adds a config option to silently ignore when a Computed Property is being set

### LINKED ISSUE #613: Data casts store computed properties
**✏️ Describe the bug**

Data casts store computed properties

**↪️ To Reproduce**

```php
class Example extends Model
{
    protected $casts = [
        'config' => ConfigData::class,
    ];
}

class ConfigData extends Data
{
    #[Computed]
    public string $test;
}

$example->save()
```

**✅ Expected behavior**

Since computed properties contain derived data they should not be stored in database

---

New method to transform for storing to database instead of `toJson`:

(https://github.com/spatie/laravel-data/blob/39647b2d1affa196cd0f5ea05cfd649a16d2d11b/src/Support/EloquentCasts/DataEloquentCast.php#L73)

It could use `->toJson()` by default

aka `trait CastableData` in addition to `trait TransformableData`

---

Note this behavior in docs:
 
https://spatie.be/docs/laravel-data/v3/as-a-data-transfer-object/computed

https://spatie.be/docs/laravel-data/v3/advanced-usage/eloquent-casting

---

Skip computed properties like hidden:

https://github.com/spatie/laravel-data/blob/39647b2d1affa196cd0f5ea05cfd649a16d2d11b/src/Transformers/DataTransformer.php#L76

```php
if ($property->computed) {
    continue;
}
```

---

Or provide another workaround

**🖥️ Versions**

Laravel: v9.52.15
Laravel Data: 3.9.1
PHP: 8.1

erikaraujo:
When creating a `Data` object, if the array contains a computed property, it'll throw an error.

This is a problem when casting a `Model` property into a `Data` if it contains a computed property.

Adding `#[Hidden]` may not be feasible for all scenarios.

In order to resolve this issue, I added a `throw_when_trying_to_set_computed_property_value` config option - which is `true` by default to avoid breaking backwards compatibility.

Setting that option as `false` will make the computed property attempted to be set to be silently ignored, allowing it to be recalculated.

----

Somewhat resolves #613 (It still stores the computed property in the DB, but doesn't error out when retrieving it).

### DIFF
--- a/config/data.php
+++ b/config/data.php
@@ -136,6 +136,13 @@ return [
     'throw_when_max_transformation_depth_reached' => true,
 
     /**
+     * When trying to set a computed property value, the package will throw an exception.
+     * You can disable this behaviour by setting this option to false, which will then just
+     * ignore the value being passed into the computed property and recalculate it.
+     */
+    'throw_when_trying_to_set_computed_property_value' => true,
+
+    /**
     * When using the `make:data` command, the package will use these settings to generate
     * the data classes. You can override these settings by passing options to the command.
     */
--- a/config/data.php
+++ b/config/data.php
@@ -17,6 +17,13 @@ return [
      */
     'features' => [
         'cast_and_transform_iterables' => false,
+
+        /**
+         * When trying to set a computed property value, the package will throw an exception.
+         * You can disable this behaviour by setting this option to true, which will then just
+         * ignore the value being passed into the computed property and recalculate it.
+         */
+        'ignore_exception_when_trying_to_set_computed_property_value' => false,
     ],
 
     /**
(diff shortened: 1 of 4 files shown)


rubenvanassche:
If I'm not mistaken the only reason to add #[Computed] to a property is to make sure this exception is thrown, you could just omit the attribute and the code should work the same?

Let me know if I'm wrong though 😅

rubenvanassche:
Move this to the features section of the config, we'll then make this default in v5.

erikaraujo:
> Move this to the features section of the config, we'll then make this default in v5.

Not sure if it needs to be default. Considering the little noise I've seen about this and that `laravel-data` is already in v4, I do believe the current way of handling `#[Computed]` is probably the best default.

But up to you.


(I don't understand why PHPStan is failing on the CI and I cannot reproduce the issue locally)

erikaraujo:
Do you want me to move the config as is to the `features` array or should I reverse it?

Like - instead of `throw_when_trying_to_set_computed_property_value`, I'd rename it to `ignore_exception_when_trying_to_set_computed_property_value` and set the default value to `false`.

rubenvanassche:
> Do you want me to move the config as is to the features array or should I reverse it?

Jup, good idea!

erikaraujo:
Done.

rubenvanassche:
Thanks!

PR #717, opened 2024-03-21
Change default cache variable

Elnadrion:
Small change:
Laravel 11 uses `CACHE_STORE` instead of old `CACHE_DRIVER` now
https://github.com/laravel/laravel/blob/11.x/.env.example#L39

### DIFF
--- a/config/data.php
+++ b/config/data.php
@@ -96,7 +96,7 @@ return [
         'enabled' => true,
         'directories' => [app_path('Data')],
         'cache' => [
-            'store' => env('CACHE_DRIVER', 'file'),
+            'store' => env('CACHE_STORE', 'file'),
             'prefix' => 'laravel-data',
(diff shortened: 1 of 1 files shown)


config/data.php
```diff
@@ -96,7 +96,7 @@
         'enabled' => true,
         'directories' => [app_path('Data')],
         'cache' => [
-            'store' => env('CACHE_DRIVER', 'file'),
+            'store' => env('CACHE_STORE', 'file'),
```
Tofandel, line 99:
```suggestion
            'store' => env('CACHE_STORE', env('CACHE_DRIVER', 'file')),
```

To be compatible with both versions

Elnadrion: Sure, fixed

rubenvanassche:
Thanks!

Issue #677, opened 2024-03-03
lazy properties not resolved correctly


edgars-vasiljevs:
**✏️ Describe the bug**
lazy properties break when empty nested object is encountered.
After null, `{id,type}` is applied to main object instead of nested one. 

It's something to do with "rollBackPartialsWhenRequired" not being called if nested object is null.

**↪️ To Reproduce**

```php
// dto
class TagData extends Data
{
    public function __construct(
        public Lazy|int $id,
        public Lazy|TagType $type, // int enum
        public Lazy|TagData|null $parent,
    ) {
    }

    public static function fromModel(Tag $tag): TagData
    {
        return new static(...[
            Lazy::create(fn () => $tag->id),
            Lazy::create(fn () => $tag->type),
            Lazy::create(fn () => $tag->parent ? TagData::from($tag->parent) : null),
        ]);
    }
}

// generate object
$data = TagData::collect($collection, DataCollection::class)
    ->include('id', 'parent.{id,type}');
```

Result:
![image](https://github.com/spatie/laravel-data/assets/456317/8ea0b658-9ca7-4583-aaec-145a6f6b2eb0)

**✅ Expected behavior**
expect null to be null and any existing objects be resolved by given include/exclude rules.

**🖥️ Versions**

Laravel: 10.10
Laravel Data: 4.2.0
PHP: 8.2.9

rubenvanassche:
Hi, could you update your test so that it doesn't use a model to show the issue. Providing an array with some default values looks like the easiest solution here.

Tofandel:
I have the same/similar issue, I'll submit a failing test as a PR

Tofandel:
@edgars-vasiljevs Can you test with my PR to see if it fixes your issue as well

rubenvanassche:
This should be fixed normally! Let me know if it is not the case

PR #720, opened 2024-03-26
Fix: nested collection lazy only includes the first element in the collection

Tofandel:
Might not be the exact same issue as #677 but definitely related

If you have a collection within a collection, including lazy keys will only include the key in the first item of the collection

### DIFF
--- a/src/Resolvers/VisibleDataFieldsResolver.php
+++ b/src/Resolvers/VisibleDataFieldsResolver.php
@@ -134,7 +134,7 @@ class VisibleDataFieldsResolver
 
             if ($nested = $exceptPartial->getNested()) {
                 try {
-                    $fields[$nested]->addExceptPartial($exceptPartial->next());
+                    $fields[$nested]->addExceptPartial((clone $exceptPartial)->next());
                 } catch (ErrorException $exception) {
                     $this->handleNonExistingNestedField($exception, PartialType::Except, $nested, $dataClass, $transformationContext);
                 }
@@ -172,7 +172,7 @@ class VisibleDataFieldsResolver
 
             if ($nested = $onlyPartial->getNested()) {
                 try {
-                    $fields[$nested]->addOnlyPartial($onlyPartial->next());
+                    $fields[$nested]?->addOnlyPartial((clone $onlyPartial)->next());
                     $onlyFields[] = $nested;
                 } catch (ErrorException $exception) {
                     $this->handleNonExistingNestedField($exception, PartialType::Only, $nested, $dataClass, $transformationContext);
@@ -279,7 +279,7 @@ class VisibleDataFieldsResolver
 
             if ($nested = $excludePartial->getNested()) {
                 try {
-                    $fields[$nested]->addExcludedPartial($excludePartial->next());
+                    $fields[$nested]?->addExcludedPartial((clone $excludePartial)->next());
                 } catch (ErrorException $exception) {
                     $this->handleNonExistingNestedField($exception, PartialType::Exclude, $nested, $dataClass, $transformationContext);
                 }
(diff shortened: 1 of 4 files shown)


tests/PartialsTest.php
```diff
@@ -225,6 +225,37 @@
...
+        ]
+    ], DataCollection::class);
+
+    expect($data->toArray())->toMatchArray([
+        ['nested' => [[]]],
+        ['nested' => [[]]],
+        ['nested' => [[]]]
+    ]);
+
+    expect(dump($data->include('nested.{artist,year}')->toArray()))
```
Tofandel, line 251: Current output

```
array:3 [
  0 => array:1 [
    "nested" => array:1 [
      0 => array:2 [
        "artist" => "Rick Astley"
        "year" => 1987
      ]
    ]
  ]
  1 => array:1 [
    "nested" => array:1 [
      0 => []
    ]
  ]
  2 => array:1 [
    "nested" => array:1 [
      0 => []
    ]
  ]
] 
```

tests/PartialsTest.php
```diff
@@ -225,6 +225,37 @@
...
+        ]
+    ], DataCollection::class);
+
+    expect($data->toArray())->toMatchArray([
+        ['nested' => [[]]],
+        ['nested' => [[]]],
+        ['nested' => [[]]]
+    ]);
+
+    expect($data->include('nested.{artist,year}')->toArray())
```
Tofandel, line 251: Current ouput

```
array:3 [
  0 => array:1 [
    "nested" => array:1 [
      0 => array:2 [
        "artist" => "Rick Astley"
        "year" => 1987
      ]
    ]
  ]
  1 => array:1 [
    "nested" => array:1 [
      0 => []
    ]
  ]
  2 => array:1 [
    "nested" => array:1 [
      0 => []
    ]
  ]
] 
```

src/Resolvers/VisibleDataFieldsResolver.php
```diff
@@ -214,6 +214,7 @@ protected function resolveIncludedFields(
                 continue;
             }
 
+            $includedPartial = clone $includedPartial;
```
Tofandel, line 217: This is what I found to fix the issue, the problem is that it's looping over the elements of the collection but all the elements of the collection get the same instance of `includedPartial` which is mutated with `includedPartial->next()` and so `$includedPartial->getNested()` returns null on the second item

src/Resolvers/VisibleDataFieldsResolver.php
```diff
@@ -230,13 +230,13 @@ protected function resolveIncludedFields(
             }
 
             if ($nested = $includedPartial->getNested()) {
+                $includedPartial = clone $includedPartial;
```
Tofandel, line 233:
```
This is what I found to fix the issue, the problem is that it's looping over the elements of the collection but all the elements of the collection get the same instance of includedPartial which is mutated with includedPartial->next() and so $includedPartial->getNested() returns null on the second item

```

Tofandel:
I'm also noting that `currentPath` in `creationContext` is also affected by a similar issue
image

rubenvanassche:
Hi @Tofandel thanks for the PR, the issue should be solved with the next version of data. I've implemented the system a bit differently.

Tofandel:
Did you use the tests in this PR and did you also fix creationContext? Can you link the commits so I can check?

rubenvanassche:
Some tests were not running that's been fixed now, when I paste your tests in the current data test suite they are green. As for creation context, could you send in a PR showing what's going on?

Tofandel:
For creation context it's just this commit https://github.com/spatie/laravel-data/pull/720/commits/f0a1a2e7870d4c7a1e595074a477ab2d1ae461d5

Basically you are going into a nested state without calling prev like in the other method where it's used, it's a small quick fix

PR #730, opened 2024-03-29
Fix: load model relations and attributes only if necessary 

### LINKED ISSUE #729: Non nullable relation collections on models are not loaded
**✏️ Describe the bug**
When declaring a Data for a model with a non nullable and non lazy collection referencing a Relation, that relation is NOT included in the final data unless the relation was explicitely loaded on the model first (using `$model->load('relation_name')`)

**↪️ To Reproduce**

```php

it('should load non nullable relations of model', function () {
    class ChapterData extends Data {
      public string $name;
    }

    class BaseData extends Data
    {
        public function __construct(
            /** @var Collection<int, ChapterData> */
            public Collection $chapters;
        ) {
        }
    }
    
    class Chapter extends Model {
    
    }
    class BaseModel extends Model {
        public function chapters() {
          return $this->hasMany(Chapter::class);
        }
    }

    $m => new BaseModel();
    dd(BaseData::from($m)->toArray()); // $m should be `['chapters' => []]` but it is `[]`
});
```

Attributes are present but present a huge performance issue, because calling `getMutatedAttributes` loads all of the available attributes of a model and not just the ones required in the DataClass, some getters can load relations and be really resource hungry but not be needed in a DataClass

Relevant code
https://github.com/spatie/laravel-data/blob/a5df67e06f1f9012261a9c14d007e81a8fd84046/src/Normalizers/ModelNormalizer.php#L32-L45

It seems a DataPipe to achieve this is required because normalizers are not passed the DataClass as a parameter (maybe they should in v5?)

Tofandel:
Fixes #729 

It also fixes the performance issue (https://github.com/spatie/laravel-data/discussions/713) of `$value->getMutatedAttributes()` which was used to load attributes into model, but the issue is it would load ALL of the attributes of the model irrespective of wether they are needed or not by the data class

Ideally the normalizer would also get the same parameters as a data pipe, but this would be a breaking change to the interface,

I don't know which makes more sense, passing the DataClass to the normalizers and only getting the needed property or doing the whole data extraction logic in the data pipe and getting rid of the model normalizer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,7 +128,7 @@ Laravel-data 4.0.0 was released 5 hours ago, time for an update!
 
 - Make ValidationPath Stringable
 - Fix PHPStan
-- Improve performance when optional peoperty exists (#612)
+- Improve performance when optional property exists (#612)
 
 ## 3.10.0 - 2023-12-01
 
--- a/src/Concerns/BaseData.php
+++ b/src/Concerns/BaseData.php
@@ -16,6 +16,7 @@ use Spatie\LaravelData\DataPipes\AuthorizedDataPipe;
 use Spatie\LaravelData\DataPipes\CastPropertiesDataPipe;
 use Spatie\LaravelData\DataPipes\DefaultValuesDataPipe;
 use Spatie\LaravelData\DataPipes\FillRouteParameterPropertiesDataPipe;
+use Spatie\LaravelData\DataPipes\LoadsModelRelationsDataPipe;
 use Spatie\LaravelData\DataPipes\MapPropertiesDataPipe;
 use Spatie\LaravelData\DataPipes\ValidatePropertiesDataPipe;
 use Spatie\LaravelData\PaginatedDataCollection;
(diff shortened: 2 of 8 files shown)


rubenvanassche:
Yeah difficult one, we need a breaking change to fix this the way I want it, gonna think about this a bit ...

Tofandel:
When I was checking, you could still pass the parameters without adding them to the interface yet and it will work without errors

And then add them in v5

linaspasv:
Any update on this?

rubenvanassche:
So I'm going to close this PR, I've made my own alternative to it which also fixes the performance problem and allows loading relations on the fly.

What's not included:

- Automatically transforming values to Optional, that's something which happens further in the pipeline, same with collections
- Automatically Lazyify values, I don't think we should do this automatically but maybe an attribute [AutoLazy] could indicate this? Would also be cool for other things than models so I certainly welcome PR's adding such a pipe! 
Otherwise I'll probably do it myself within a few months
- Support for counting relations is present at the moment, probably something to think about in the future

Thanks for your initial PR, it was quite handy to have some starting point!

Tofandel:
> Automatically Lazyify values, I don't think we should do this automatically but maybe an attribute [AutoLazy] could indicate 

> this? Would also be cool for other things than models so I certainly welcome PR's adding such a pipe!
> Otherwise I'll probably do it myself within a few months

The lazyfication needs to happen before the relation is actually loaded otherwise it doesn't provide any advantage and the queries would still happen, so each pipe needs to implement its own lazyfication logic, to me model relations is the only thing that really benefits from being created as a lazy if possible

PR #736, opened 2024-04-04
Fix for the use of built-in class names in Collection annotations

27pchrisl:
Hi, found this when upgrading from v3 to v4.

In v3 you could do

```php
#[DataCollectionOf(Error::class)]
```

which would correctly pick up a class with the right namespace, even if its basename matches a builtin like `\Error`.

In v4 the same syntax is:

```php
/** @var Collection<Error> */
public Collection $errors;
```

But due to the order in which the `DataIterableAnnotationReader` determines the correct class, it discovers `\Error` before `App\Stuff\Error` or similar.

This PR reorders the resolution code so that it finds properly qualified classes before builtins.

### DIFF
--- a/src/Support/Annotations/DataIterableAnnotationReader.php
+++ b/src/Support/Annotations/DataIterableAnnotationReader.php
@@ -170,19 +170,19 @@ class DataIterableAnnotationReader
             ];
         }
 
-        if (class_exists($class)) {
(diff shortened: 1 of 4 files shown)


27pchrisl:
Note this branch needs rebasing when tests pass again on `main`

rubenvanassche:
Looking great, thanks!

PR #741, opened 2024-04-08
Abstract data object as collection

onursimsek:
With this feature, abstract data objects can be used as collections.

Ps: [old comment](https://github.com/spatie/laravel-data/pull/526#issuecomment-1676603238)

### DIFF
--- a/docs/advanced-usage/eloquent-casting.md
+++ b/docs/advanced-usage/eloquent-casting.md
@@ -115,6 +115,19 @@ The child data object value of the model will be stored in the database as a JSO
 
 When retrieving the model, the data object will be instantiated based on the `type` key in the JSON string.
 
+#### Abstract data object with collection
+
+You can use with collection.
+
+```php
+class Record extends Model
+{
+    protected $casts = [
+        'configs' => DataCollection::class . ':' . RecordConfig::class,
+    ];
+}
+```
+
 #### Abstract data class morphs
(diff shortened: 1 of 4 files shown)


rubenvanassche:
Thanks!

ManukMinasyan:
Hi @rubenvanassche @onursimsek,

I'm encountering an error when trying to implement polymorphic expressions in `FilterGroupContentData`. 

When trying to use `FieldExpression` or `RelationshipExpression` within the `expressions` data collection, I'm getting the following error:

```
Cannot instantiate abstract class DataLens\Data\Filters\FilterExpression
```

Here's my current implementation:

```php
abstract class FilterExpression extends Data
{
    public function __construct(
        public string $type,
    ) {}
}

//

class FilterGroupContentData extends Data
{
    public function __construct(
        /*  @var FieldExpression|RelationshipExpression */
        #[DataCollectionOf(FilterExpression::class)]
        public DataCollection $expressions,
    )
    {
    }
}
```

I attempted to use it inside a data object but haven't been successful. Could you please advise on how to properly integrate a solution that supports polymorphic expressions? I need to handle multiple `FilterGroupContentData` instances, each containing various expressions of different types (`FieldExpression` or `RelationshipExpression`).

Do you have any recommendations for working with abstract classes for this specific use case?

Thanks for your help!

PR #747, opened 2024-04-14
Feature: private/protected data properties

trevorpe:
## Description

This PR attempts to add the ability to mark private/protected properties as data properties. This would allow using Laravel Data functionality within more traditional classes that encapsulate their data and only expose it via public getter/setter methods.

Data properties are still determined in the same way and transformations can be applied even to values obtained from getter calls.

## Questions

1. Does applying Laravel Data traits to other, non-DTO-like classes stray too far from the original intent of the package?
1. Are there other places where data is get/set that need to be updated?

### DIFF
--- a/src/Attributes/DataProperty.php
+++ b/src/Attributes/DataProperty.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Spatie\LaravelData\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+class DataProperty
+{
+    public function __construct(public readonly ?string $getter = null, public readonly ?string $setter = null)
+    {
+    }
+}
--- a/src/Resolvers/DataFromArrayResolver.php
+++ b/src/Resolvers/DataFromArrayResolver.php
@@ -61,7 +61,11 @@ class DataFromArrayResolver
                 continue; // Ignore the value being passed into the computed property and let it be recalculated
             }
 
-            $data->{$property->name} = $properties[$property->name];
+            if ($setter = $property->setter) {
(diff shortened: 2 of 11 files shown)


rubenvanassche:
Hi @trevorpe 

Thanks for your PR, at the moment we're going to keep data properties only public since I don't see a use case for private and protected properties in data objects.

trevorpe:
Hi @rubenvanassche ,

Thanks for taking a look. I've given more context below in case you'd like to explore this use case at all, but I leave it up to you. The purpose and intent of the library may be to stick closely to DTO-like classes, which is understandable and would most likely exclude this use case. Perhaps another library could extend this one for this purpose.

---
For more context: my thoughts were that this library does many things really well, including representing the schema of a data object in a friendly enough way to work with and has tight integration with Laravel.

We use Doctrine ORM in our Laravel project. Doctrine entity classes are just plain PHP classes, usually with private/protected state exposed via traditional getters and setters. We often want to serialize those instances to JSON or similar. We can manually implement a `toArray()`, of course. One could also create dedicated `Data` classes for it and/or the various views of that data. This requires a bit of data wrangling, however.

The traits in laravel-data handles a lot of the common parts of serialization, but explicitly depends on variables being declared public, which makes a lot of sense for DTO-type classes. With a slight tweak, I thought it might be able to make the trait usable by more traditional classes with private state.

rubenvanassche:
Hmm I get your point, at the moment I want to keep the package simple and not add yet other branches of how data can flow since it is already complicated and we still have some bugs here and there.

In a future version we'll maybe make it possible to add this but at the moment I'm going to keep it restricted to public properties only.

PR #749, opened 2024-04-15
Fix creation context issue

Tofandel:
Follow up of #720 in which I also fixed the creation context path being invalid when going into nested collections

Also added the test of #720 which was not added to main to avoid a future regression

### DIFF
--- a/src/DataPipes/CastPropertiesDataPipe.php
+++ b/src/DataPipes/CastPropertiesDataPipe.php
@@ -87,9 +87,13 @@ class CastPropertiesDataPipe implements DataPipe
         ) {
             $context = $creationContext->next($property->type->dataClass, $property->name);
 
-            return $property->type->kind->isDataObject()
+            $data = $property->type->kind->isDataObject()
                 ? $context->from($value)
                 : $context->collect($value, $property->type->iterableClass);
+
+            $creationContext->previous();
+
+            return $data;
         }
 
         if (
--- a/src/DataPipes/CastPropertiesDataPipe.php
+++ b/src/DataPipes/CastPropertiesDataPipe.php
@@ -94,6 +94,7 @@ class CastPropertiesDataPipe implements DataPipe
                   : $context->collect($value, $property->type->iterableClass);
 
               $creationContext->previous();
+                return $data;
             } catch (CannotCreateData) {
                 return $value;
             }
--- a/src/DataPipes/CastPropertiesDataPipe.php
+++ b/src/DataPipes/CastPropertiesDataPipe.php
@@ -87,15 +87,15 @@ class CastPropertiesDataPipe implements DataPipe
             || $property->type->kind->isDataCollectable()
         ) {
             try {
-              $context = $creationContext->next($property->type->dataClass, $property->name);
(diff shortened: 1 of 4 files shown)


tests/CreationTest.php
```diff
@@ -1068,6 +1072,63 @@ public function __invoke(SimpleData $data)
...
+    }
+
+    class SimpleDataWithTestPipe extends SimpleData
+    {
+        public static function pipeline(): DataPipeline
+        {
+            return parent::pipeline()
+                ->through(TestNestedDataPipe::class);
+        }
+    }
```
Tofandel, line 1106:
```suggestion

    class SimpleDataWithTestPipe extends SimpleData
    {
        public static function pipeline(): DataPipeline
        {
            return parent::pipeline()
                ->through(TestDataPipe::class);
        }
    }
```

src/DataPipes/CastPropertiesDataPipe.php
```diff
@@ -86,12 +86,14 @@ protected function cast(
...
-                    ? $context->from($value)
-                    : $context->collect($value, $property->type->iterableClass);
+              $context = $creationContext->next($property->type->dataClass, $property->name);
+              
+              $data = $property->type->kind->isDataObject()
+                  ? $context->from($value)
+                  : $context->collect($value, $property->type->iterableClass);
+
+              $creationContext->previous();
             } catch (CannotCreateData) {
```
Tofandel, line 97:
```suggestion
                return $data;
            } catch (CannotCreateData) {
```

src/DataPipes/CastPropertiesDataPipe.php
```diff
@@ -86,12 +86,15 @@ protected function cast(
...
-                    ? $context->from($value)
-                    : $context->collect($value, $property->type->iterableClass);
+              $context = $creationContext->next($property->type->dataClass, $property->name);
+              
+              $data = $property->type->kind->isDataObject()
+                  ? $context->from($value)
+                  : $context->collect($value, $property->type->iterableClass);
+
+              $creationContext->previous();
+                return $data;
```
Tofandel, line 97:
```suggestion
              return $data;
```

src/DataPipes/CastPropertiesDataPipe.php
```diff
@@ -86,12 +86,16 @@ protected function cast(
...
-                    : $context->collect($value, $property->type->iterableClass);
+              $context = $creationContext->next($property->type->dataClass, $property->name);
+              
+              $data = $property->type->kind->isDataObject()
+                  ? $context->from($value)
+                  : $context->collect($value, $property->type->iterableClass);
+
+              $creationContext->previous();
+
+              return $data;
```
Tofandel, line 98:
```suggestion
                $context = $creationContext->next($property->type->dataClass, $property->name);
              
                $data = $property->type->kind->isDataObject()
                    ? $context->from($value)
                    : $context->collect($value, $property->type->iterableClass);

                $creationContext->previous();

                return $data;
```

rubenvanassche:
Thanks!

PR #773, opened 2024-05-08
Avoid loading already loaded relations and allow loading non studly relation names

Tofandel:
`load` would load relations even if they were already loaded in the model

But we might not even need to do that explicitely because `$this->model->{$studlyName}` would also load it

Added a fix for #787 as well

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -46,4 +46,4 @@ jobs:
           composer update --${{ matrix.stability }} --prefer-dist --no-interaction
 
       - name: Execute tests
-        run: vendor/bin/pest
+        run: vendor/bin/pest --enforce-time-limit --fail-on-risky
--- a/src/Normalizers/Normalized/NormalizedModel.php
+++ b/src/Normalizers/Normalized/NormalizedModel.php
@@ -69,18 +69,23 @@ class NormalizedModel implements Normalized
             return $this->properties[$name] = $this->model->getAttribute($name);
         }
 
-        if (! $dataProperty->attributes->contains(fn (object $attribute) => $attribute::class === LoadRelation::class)) {
-            return UnknownProperty::create();
-        }
-
         $studlyName = Str::studly($name);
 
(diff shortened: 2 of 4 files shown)


Tofandel:
`$studlyName = Str::studly($name);` is also not the correct way to search for a relation, it can both be studly or not (see #730)

Tofandel:
Also worth noting that because of this change, the following will work whether the SnakeCaseMapper is used or not

```
        #[DataCollectionOf(FakeNestedModelData::class), MapInputName(SnakeCaseMapper::class)]
        public array|Optional $fakeNestedModels;
```

sebastiaanluca:
I was checking the new code and noticed the nested eager load (https://github.com/spatie/laravel-data/compare/4.5.1...4.6.0#diff-5a676e204bcb067cd1bf62ef73416609c339f8e16479ab7b38021e9799c2d607R82). Is it possible to remove this line completely?

If you have a collection of models it'll create an n+1 relation loading issue when creating a property from a relation. The issue remains even when using `loadMissing()`. IMO the normalizer class should not be in charge of eager loading anything, that's up to the user, higher up in the controller or elsewhere.

I was going to create a PR for it, but maybe we can include it in this one? I believe I also highlighted the issue a while ago (in this repo or another) where a newly introduced nested eager load triggered a ton of queries in our app and we didn't catch it for a long time because it was included in a minor or patch release. The n+1 exception isn't thrown because the relation isn't lazy loaded.

Tofandel:
> it'll create an n+1 relation loading issue when creating a property from a relation

That doesn't really matter this only happens with a special attribute and not by default so you already have to be explicit about it, of course if you have 4 nested relations with 100 of models you'll want to load them yourself, but for 10 models with 1 level or 2 of nesting, it's not worth removing this feature

We could later improve it by using Lazy and having the tree of relation in the context to load all of the needed relations at the end of the traversal before the relations are actually loaded one by one

rubenvanassche:
Thanks!

Issue #771, opened 2024-05-08
Incorrect injection of routing parameters when using MapInputName


guiqibusixin:
**✏️ Describe the bug**
Incorrect injection of routing parameters when using MapInputName.

**My English is not very good. If the expression is not clear, I feel sorry**

**↪️ To Reproduce**

```php

it('incorrect injection of routing parameters when using property name mapping', function () {
    #[MapInputName(SnakeCaseMapper::class)]
    #[MapOutputName(SnakeCaseMapper::class)]
    class CreateRequest extends Data
    {
        #[FromRouteParameter('user_id')]
        public int $userId;
    }
});
```

**✅ Expected behavior**
Can correctly inject routing parameters when using MapInputName

**Reason**
In the FillRouteParameterPropertiesDataPipe class, only the property name used is injected without considering the property mapping name.
My temporary solution now is to replace the class FillRouteParameterPropertiesDataPipe, with the following code
```php
public function handle(
        mixed $payload,
        DataClass $class,
        array $properties,
        CreationContext $creationContext
    ): array {
        if (! $payload instanceof Request) {
            return $properties;
        }

        foreach ($class->properties as $dataProperty) {
            $attribute = $dataProperty->attributes->first(
                fn (object $attribute) => $attribute instanceof FromRouteParameter || $attribute instanceof FromRouteParameterProperty
            );

            if ($attribute === null) {
                continue;
            }

            // if inputMappedName exists, use it first
            $name = $dataProperty->inputMappedName ?: $dataProperty->name;
            if (! $attribute->replaceWhenPresentInBody && array_key_exists($name, $properties)) {
                continue;
            }

            $parameter = $payload->route($attribute->routeParameter);

            if ($parameter === null) {
                continue;
            }

            $properties[$name] = $this->resolveValue($dataProperty, $attribute, $parameter);

            // keep the original property name
            if ($name !== $dataProperty->name) {
                $properties[$dataProperty->name] = $properties[$name];
            }
        }

        return $properties;
    }
```

**🖥️ Versions**

Laravel: 10.48.4
Laravel Data: 4.4.1
PHP: 8.1.27

rubenvanassche:
Can you PR this? Seems like a valid solution to me!

rubenvanassche:
Fixed with PR

PR #775, opened 2024-05-08
Fix routing parameters filled in incorrectly using mapping property names

guiqibusixin:
#771

### DIFF
--- a/src/DataPipes/FillRouteParameterPropertiesDataPipe.php
+++ b/src/DataPipes/FillRouteParameterPropertiesDataPipe.php
@@ -31,7 +31,9 @@ class FillRouteParameterPropertiesDataPipe implements DataPipe
                 continue;
             }
 
-            if (! $attribute->replaceWhenPresentInBody && array_key_exists($dataProperty->name, $properties)) {
+            // if inputMappedName exists, use it first
+            $name = $dataProperty->inputMappedName ?: $dataProperty->name;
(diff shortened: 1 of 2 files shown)


src/DataPipes/FillRouteParameterPropertiesDataPipe.php
```diff
@@ -41,7 +43,12 @@ public function handle(
                 continue;
             }
 
-            $properties[$dataProperty->name] = $this->resolveValue($dataProperty, $attribute, $parameter);
+            $properties[$name] = $this->resolveValue($dataProperty, $attribute, $parameter);
+
+            // keep the original property name
+            if ($name !== $dataProperty->name) {
+                $properties[$dataProperty->name] = $properties[$name];
+            }
```
Tofandel, line 51: Should probably just be

```suggestion
$properties[$dataProperty->outputMappedName ?? $dataProperty->name] = $this->resolveValue($dataProperty, $attribute, $parameter);
```

Instead of the next line

rubenvanassche: Yeah that would seem the best solution, the problem is that validation now chooses the mapped name over the original name (this is a bug I need to fix but it is more complicated then I initially thought) while the data object is created from the original name ...

Something I need time to look at but is more complicated than you would expect since we need to know if a name was mapped for outputting errors.

rubenvanassche:
Thanks!

PR #781, opened 2024-05-28
Fix dataCastUsing method signature

shankhadevpadam:
The dataCastUsing method doesn't follow the Cast interface signature.


rubenvanassche:
Thanks!

PR #808, opened 2024-07-02
Fix Parsing of Optional Types Annotations in DataIterableAnnotation

clementbirkle:
This PR fixes an issue where DataIterableAnnotation couldn't correctly identify data types with optional parameters. 

Previously, the test for "is data" returned:

```
// Previously failed
@param \Spatie\LaravelData\DataCollection<\Spatie\LaravelData\Tests\Fakes\SimpleData>|null $paramK (false)

// Previously worked
@param null|\Spatie\LaravelData\DataCollection<\Spatie\LaravelData\Tests\Fakes\SimpleData> $paramK (true)

```

Now, the test returns `true` for both cases, ensuring accurate data type identification regardless of optional parameters like `|null`.


rubenvanassche:
Cool, thanks!

Issue #681, opened 2024-03-05
The behavior of the RequiredRuleInferrer is confusing


alexrififi:
**✏️ Describe the bug**
If you make a property of a Data of type `bool`, then the Required auto rule will not be applied to it.

**↪️ To Reproduce**

```php
use Illuminate\Validation\ValidationException;
use Spatie\LaravelData\Attributes\Validation\Required;
use Spatie\LaravelData\Data;
use Spatie\LaravelData\Exceptions\CannotCreateData;

it('successfully validated', function () {
    class SomeData1 extends Data
    {
        public function __construct(
            public string $just_string,
            public bool $i_am_bool,
        ) {
        }
    }

    $payload = ['just_string' => fake()->text()];
    $validated = SomeData1::validate($payload);
    expect($validated['just_string'])->toEqual($payload['just_string']);
});

it('validated, but not created', function () {
    class SomeData2 extends Data
    {
        public function __construct(
            public string $just_string,
            public bool $i_am_bool,
        ) {
        }
    }

    SomeData2::validateAndCreate(['just_string' => fake()->text()]);
})->throws(CannotCreateData::class);

it('fail validation', function () {
    class SomeData3 extends Data
    {
        public function __construct(
            public string $just_string,
            #[Required]
            public bool $i_am_bool,
        ) {
        }
    }

    SomeData3::validateAndCreate(['just_string' => fake()->text()]);
})->throws(ValidationException::class);
```

**✅ Expected behavior**

I expect that I won't have to manually add the `Required` attribute

**🖥️ Versions**

Laravel: 10.46.0
Laravel Data: 4.2.0
PHP: 8.3.1

---

## The problem is here

https://github.com/spatie/laravel-data/blob/6e20d6f191b4260e64738bd299d2aa50dae42297/src/RuleInferrers/RequiredRuleInferrer.php#L38-L40

rubenvanassche:
Yeah this has always been like this and has to do with checkboxes, they do not send a value when not being checked and thus validation would fail since the key is missing.

That being said,  the behavior looks weird after all those years. Data isn't a request to data solution anymore, we're powering event sourcing objects with it, not even using a request in between. These days when using Inertia with react/vue, the boolean fields will be present from the start so I agree that this change we've made in the past now feels weird.

The problem is, that just adding the required rule is a massive breaking change since a lot of validation which was previously not throwing error will do now.

[...]

Issue #647, opened 2024-01-24
Cannot use multiple requiring rules


striker4150:
**✏️ Describe the bug**
Adding multiple requiring rules to a property/parameter only results in the last added requiring rule being applied. For example, if a `RequiredUnless` rule is used alongside a `RequiredWith` rule, only the `RequiredWith` rule will be applied (assuming that the `RequiredWith` rule is last).

The offending lines seem to be:
https://github.com/spatie/laravel-data/blob/324970e910414b92b43a239843efd06fa51090e9/src/Support/Validation/PropertyRules.php#L40-L44

**↪️ To Reproduce**
```php

it('cannot add multiple requiring rules', function () {
    class BaseData extends Data
    {
        public function __construct(
            #[RequiredUnless('requires_a', false), RequiredWith('b')]
            public string $a,
            public string $b,
            public bool $requires_a,
        ) {
        }
    }

    dd(BaseData::getValidationRules([]));
});
```

**✅ Expected behavior**
In the scenario above, I expect the `a` property to have the requiring rules `['required_unless:requires_a,false', 'required_with:b']`. However, the `a` property only has the `required_with:b`. Laravel *does* offer the ability to apply multiple requiring rules at once, as demonstrated here:

```php
Validator::make(
  [
    "a" => "a",
    "b" => "b",
    "requires_a" => false
  ],
  [
    "a" => ["required_unless:requires_a,false", "required_with:b"]
  ]
)->passes();
// Returns true

Validator::make(
  [
    "a" => "",
    "b" => "b",
    "requires_a" => true
  ],
  [
    "a" => ["required_unless:requires_a,false", "required_with:b"]
  ]
)->passes();
// Returns false (it's required with b AND requires_a is true)

Validator::make(
  [
    "a" => "",
    "b" => "b",
    "requires_a" => false,
  ],
  [
    "a" => ["required_unless:requires_a,false", "required_with:b"]
  ]
)->passes();
// Returns false (it's required with b)

Validator::make(
  [
    "a" => "",
    "b" => "",
    "requires_a" => true,
  ],
  [
    "a" => ["required_unless:requires_a,false", "required_with:b"]
  ]
)->passes();
// Returns false (requires_a is true)

Validator::make(
  [
    "a" => "",
    "b" => "",
    "requires_a" => false
  ],
  [
    "a" => ["required_unless:requires_a,false", "required_with:b"]
  ]
)->passes();
// Returns true
```

[...]

PR #813, opened 2024-07-06
Add require_rule_for_booleans option

clementbirkle:
This PR concerns this issue: https://github.com/spatie/laravel-data/issues/681

### DIFF
--- a/config/data.php
+++ b/config/data.php
@@ -67,6 +67,11 @@ return [
     ],
 
     /**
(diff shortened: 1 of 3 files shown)


rubenvanassche:
Thanks for your PR, at the moment I'm not going to merge it since we want a more structural solution with a new RuleInferrer as said in #681 which also solves #647.

Although the solution might work, it is a bit hacky and that's something I want to avoid because it creates too many execution paths which we should cover.

PR #814, opened 2024-07-06
Add integer_rule_for_integers option

clementbirkle:
By default, integer attributes validation accepts both float and integer values. In some cases, it can be useful to only accept integers.

### DIFF
--- a/config/data.php
+++ b/config/data.php
@@ -67,6 +67,11 @@ return [
     ],
 
     /**
+     * By default, integer attributes validation accepts both float and integer values. In some cases, it can be useful to only accept integers.
+     */
+    'integer_rule_for_integers' => false,
+
+    /**
      * Normalizers return an array representation of the payload, or null if
      * it cannot normalize the payload. The normalizers below are used for
      * every data object, unless overridden in a specific data object class.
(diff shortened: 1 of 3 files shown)


rubenvanassche:
Thanks for your PR but I think it is up to the user to add this manually using an attribute or write their own RuleInferrer, so I'm not going to merge it.

MannikJ:
At the very least, the [docs](https://spatie.be/docs/laravel-data/v4/validation/auto-rule-inferring) should be corrected to reflect the current behavior. Right now, it states that an `int` attribute will result in an `integer` validation rule, which isn’t true.

The actual logic in `BuiltInTypesRuleInferrer` clearly adds only a `numeric` rule for `int`:

```php
if ($property->type->type->acceptsType('int')) {
$rules->add(new Numeric());
}
```

This causes confusing behavior — values like `5.5` will pass validation and be silently coerced to `5`, which undermines the intention of explicitly using `int` as a type.

While I understand the idea of adding a config option (as this PR proposes), I think that a config-wide toggle might too coarse.
A better approach fitting to the style of this package might be to introduce a custom attribute — something like:

```php
#[StrictInteger]
public int $foo;
```

Or even a class-level override:

```php
#[UseStrictIntegers]
class MyData extends DataTransferObject { ... }
```

This would:

Keep the existing `numeric` inference as the default (non-breaking),

Let consumers opt-in to strict `integer` rules where they actually want them,

Avoid global side effects for teams or projects that rely on `numeric` coercion.

PR #819, opened 2024-07-09
Give an option to set a date at midnight

PhilippeThouvenot:
Set a casted date to midgnit, avoir current time in date when an input which doesn't have time Ex: Y-m-d

#[WithCast(DateTimeInterfaceCast::class, setMidnight: true)]


rubenvanassche:
Mhh I think it is up to the user to change the data after the data creation, feels a bit sketchy to add an option which doesn't exist in PHP.

PR #820, opened 2024-07-09
Support TypeScript Hidden Properties

hfoletto:
When using the `typescript-transformer` package, it's possible to describe properties as `Hidden`: https://spatie.be/docs/typescript-transformer/v2/dtos/typing-properties#content-hidden-types

That doesn't work at the moment when using Data classes from `laravel-data` because the `DataTypeScriptTransformer` doesn't look for that attribute.

This PR address this issue.


rubenvanassche:
Thanks!

PR #841, opened 2024-08-14
Passing an enum to caster handle without an error

lkmadushan:
Allow to pass an enum to `Structure::validateAndCreate` method.

Previously below code didn't work.

```
Structure::validateAndCreate([
    'enum' => DummyBackedEnum::FOO
])
```

Some libraries (ex: [Lighthouse PHP](https://lighthouse-php.com/)) which is a GraphQL implementation, sends Enum types through the request. I think we don't need to send the Enum string value again to the validate method. It should accept Enum types also.

### DIFF
--- a/src/Casts/EnumCast.php
+++ b/src/Casts/EnumCast.php
@@ -36,6 +36,10 @@ class EnumCast implements Cast, IterableItemCast
             return Uncastable::create();
         }
 
+        if ($value instanceof $type) {
+            return $value;
+        }
+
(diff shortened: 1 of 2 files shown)


clementbirkle:
I have the same issue with attributes that are collections of enums.

I get an error like: `Object of class App\Enums\MyEnum could not be converted to string`.

I think/hope that this PR should fix it as well.

rubenvanassche:
Thanks!

PR #842, opened 2024-08-14
Passing date objects to caster handle without an error

lkmadushan:
Same as: https://github.com/spatie/laravel-data/pull/841


rubenvanassche:
Thanks!

PR #846, opened 2024-08-17
feat: Add default name mappers

przemyslaw-przylucki:
The goal of this PR is to allow for specifying the default mapping of keys for data objects.

It's an opt-in feature that can be overriden by inline attribute whenever needed

### DIFF
--- a/config/data.php
+++ b/config/data.php
@@ -1,6 +1,7 @@
 <?php
 
 use Illuminate\Support\Enumerable;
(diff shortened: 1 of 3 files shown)


timmaier:
This would be a nice config feature related to discussion https://github.com/spatie/laravel-data/discussions/750

rubenvanassche:
Cool feature, thanks!

PR #848, opened 2024-08-19
Add the ability to optionally merge automatically inferred rules with manual rules

CWAscend:
## 💪 Motivation

I would like the ability to use a combination of Attribute rules and manual rules that are defined within the static `rules` method in a `Data` object. For me, this will massively clean up my rules method when I have to reach for it, such as which I might be reaching for a `callable` rule, or using the `Rule::in()` rule where the array is dynamic - just to name some examples.

## 🛠 Changes

- Added a new `Spatie\LaravelData\Attributes\MergeRules` attribute
- This will merge the rules from the rules method rather than overwriting them
- Example usage can be found in the test Data class here: `tests/Fakes/DataWithMergedRuleset.php`

## 🧪 Testing

- A new test has been created: `it will merge validation rules`, which ensures both validation rules from the Attributes and the rules method are picked up.
- A new test has been created: `it will merge validation rules using string rules` which ensures the validation rules can be a piped string of rules instead of an array of files for each attribute

## 🧙 Reminders (Author)

- [x] Have you read through your own diff? 👀
- [x] Are all tests passing? 🧪


CheesyTech:
Up

src/Resolvers/DataValidationRulesResolver.php
```diff
@@ -240,9 +241,19 @@ protected function resolveOverwrittenRules(
...
         );
 
-        $overwrittenRules = app()->call([$class->name, 'rules'], ['context' => $validationContext]);
+        $manualRules = app()->call([$class->name, 'rules'], ['context' => $validationContext]);
 
-        foreach ($overwrittenRules as $key => $rules) {
+        if ($this->shouldMergeRules($class)) {
+            $manualRules = collect($manualRules)->map(
+                fn (string|array $rules) => is_array($rules) ? $rules : explode('|', $rules)
+            )->all();
```
rubenvanassche, line 249: Why not use the `RuleDenormalizer` here?

Also: when we add an attribute not to validate the property it will still be validated we don't want that.

I think a better solution would be to implement the logic within the loop where addRules isn't called since it replaces them but rather rules are merged over there (or not if the property shouldn"t be validated).

src/Resolvers/DataValidationRulesResolver.php
```diff
@@ -278,4 +289,9 @@ protected function inferRulesForDataProperty(
             $path
         );
     }
+
+    protected function shouldMergeRules(DataClass $class): bool
+    {
+        return $class->attributes->contains(fn (object $attribute) => $attribute::class === MergeRules::class);
```
rubenvanassche, line 295: No need for an extra function here, just inline the variable

src/Resolvers/DataValidationRulesResolver.php
```diff
@@ -223,7 +224,7 @@ protected function resolveToplevelRules(
     }
 
 
-    protected function resolveOverwrittenRules(
+    protected function resolveRules(
```
rubenvanassche, line 227: Let us rename this to appendOverwrittenRules, resolveRules is too general.

src/Resolvers/DataValidationRulesResolver.php
```diff
@@ -240,9 +241,19 @@ protected function resolveOverwrittenRules(
             $path
         );
 
-        $overwrittenRules = app()->call([$class->name, 'rules'], ['context' => $validationContext]);
+        $manualRules = app()->call([$class->name, 'rules'], ['context' => $validationContext]);
```
rubenvanassche, line 244: Let's keep this as overwrittenRules.

tests/Fakes/DataWithMergedRuleset.php
```diff
@@ -0,0 +1,24 @@
...
+        #[Max(10)]
+        public string $first_name,
+    ) {
+    }
+
+    public static function rules(): array
+    {
+        return [
+            'first_name' => ['min:2']
+        ];
```
rubenvanassche, line 22: Inline these classes within the tests, that makes the fakes directory a bit less heavy.

rubenvanassche (changes requested):
Good idea! Some changes required in order to get it merged.

src/Resolvers/DataValidationRulesResolver.php
CWAscend: I've moved this to wher you have suggested, and added a `merge` function on the `DataRules`

src/Resolvers/DataValidationRulesResolver.php
CWAscend: Have removed this

src/Resolvers/DataValidationRulesResolver.php
CWAscend: I've reverted this back to `resolveOverwrittenRules` - I agree `resolveRules` is too generic, but `appendOverwrittenRules` could suggest this method is always merging rather than replacing.

Let me know what you think :)

src/Resolvers/DataValidationRulesResolver.php
CWAscend: Have sorted this

tests/Fakes/DataWithMergedRuleset.php
CWAscend: Have sorted this, and have added a new test to check validation rules are skipped using `WithoutValidation`

CWAscend:
@rubenvanassche this is ready to be re-reviewed 👍

rubenvanassche:
Thanks!

PR #849, opened 2024-08-19
Allow to specify attribute name in type class

azzumed:
Useful change to support resolving attribute name from value-object like classes


azzumed:
@rubenvanassche what you think about merging it?

rubenvanassche:
Mhh at the moment I'm not going to merge this, if such a functionality should exists I think it should be implemented within Laravel and not an external package just to keep validation as close to Laravel as possible.

PR #858, opened 2024-09-07
Updated the wrong validation page link to the right one.

skdishansachin:
I updated the wrong validation page link to the right one.


rubenvanassche:
Thanks!

PR #876, opened 2024-10-02
Add array type annotations on Cast interfaces

adriaanzon:
Add type annotations for the `$properties` array on the public Cast interfaces, making it clearer what kind of array to expect, and preventing phpstan errors like these:

```
 ------ -------------------------------------------------------------------------------------------------------------------------
  Line   app/Data/Casts/MyCast.php
 ------ -------------------------------------------------------------------------------------------------------------------------
  22     Method App\Data\Casts\MyCast::cast() has parameter $properties with no value type specified in iterable type array.
         💡 See: https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type
 ------ -------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/src/Casts/Cast.php
+++ b/src/Casts/Cast.php
@@ -7,5 +7,8 @@ use Spatie\LaravelData\Support\DataProperty;
 
 interface Cast
 {
+    /**
+     * @param array<string, mixed> $properties
+     */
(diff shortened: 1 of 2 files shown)


rubenvanassche:
Thanks!

Issue #870, opened 2024-09-24
Compatibility with spatie/laravel-translatable is broken


alexrififi:
> In spatie/laravel-translatable translatable attributes are stored as an array and converted to a string [like this](https://github.com/spatie/laravel-translatable/blob/main/src/HasTranslations.php#L30-L37)

**✏️ Describe the bug**

Version 4.6.0 introduced a new `NormalizedModel`.

In which, after some time, the [method for obtaining model attributes was changed](https://github.com/spatie/laravel-data/commit/b7e10d345c059ff1de04168ec1d75b65157ac5a1). And that broke the compatibility.

Because of this, translatable attributes began to come to the data class as arrays, instead of strings.

An we got this error

> Argument ... must be of type string, array given

We can write workarounds to avoid problems.

But... It looks terrible 

```php
    // trait HasTranslations
    public function attributesToArray(): array
    {
        $result = parent::attributesToArray();

        [, $caller] = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2);
        if ($caller['class'] === NormalizedModel::class) {
            foreach ($this->getTranslatableAttributes() as $attribute) {
                $result[$attribute] = $this->getAttributeValue($attribute);
            }
        }

        return $result;
    }
```

I can create a custom `ModelSerializer`, but then I'll have to completely repeat `NormalizedModel` and make changes there.

The simplest solution is to make the `$properties` property of the `NormalizedModel` class **public** and then I can use my custom normalizer right after `ModelNormalizer` and cast the translatable properties to a string. Also adding a setter as mentioned in #803 would also solve the problem...

Please tell me, is this possible? Or maybe there is another way to solve the problem?

This problem was also mentioned in #851 

Thank you for reading.

**↪️ To Reproduce**

Unfortunately there is no test to show the problem

**🖥️ Versions**

Laravel: 10.48.22
Laravel Data: 4.9.0
PHP: 8.3

Tofandel:
It did break stuff, I'm just testing this and now it's the opposite of this issue in another translation lib https://github.com/Astrotomic/laravel-translatable/blob/main/src/Translatable/Translatable.php#L143

It uses `getAttribute` instead of `getAttributeValue` so now we don't retrieve anything when we try to get a translated attribute

PR #883, opened 2024-10-10
Fix retrieval of property to work for Astrotomic/translatable

Tofandel:
As seen in #870 the move to `getAttributeValue` bypasses the overriden method https://github.com/Astrotomic/laravel-translatable/edit/main/src/Translatable/Translatable.php#L143

This PR goes back to using `getAttribute` but after already checking if it's a relation (because getAttribute loads the relation otherwise, which we don't want), like this both spatie/translatable and astrotomic/translatable will work with this package

### DIFF
--- a/src/Normalizers/Normalized/NormalizedModel.php
+++ b/src/Normalizers/Normalized/NormalizedModel.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\LaravelData\Normalizers\Normalized;
 
+use Illuminate\Database\Eloquent\MissingAttributeException;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Str;
 use ReflectionProperty;
@@ -23,11 +24,9 @@ class NormalizedModel implements Normalized
 
     public function getProperty(string $name, DataProperty $dataProperty): mixed
     {
-        $propertyName = $this->model::$snakeAttributes ? Str::snake($name) : $name;
-
-        $value = array_key_exists($propertyName, $this->properties)
-            ? $this->properties[$propertyName]
-            : $this->fetchNewProperty($propertyName, $dataProperty);
+        $value = array_key_exists($name, $this->properties)
+            ? $this->properties[$name]
+            : $this->fetchNewProperty($name, $dataProperty);
 
         if ($value === null && ! $dataProperty->type->isNullable) {
             return UnknownProperty::create();
@@ -38,25 +37,23 @@ class NormalizedModel implements Normalized
 
     protected function fetchNewProperty(string $name, DataProperty $dataProperty): mixed
     {
-        if ($this->hasModelAttribute($name)) {
-            return $this->properties[$name] = $this->model->getAttributeValue($name);
-        }
-
-        $camelName = Str::camel($name);
-
         if ($dataProperty->attributes->contains(fn (object $attribute) => $attribute::class === LoadRelation::class)) {
             if (method_exists($this->model, $name)) {
                 $this->model->loadMissing($name);
-            } elseif (method_exists($this->model, $camelName)) {
-                $this->model->loadMissing($camelName);
             }
         }
 
         if ($this->model->relationLoaded($name)) {
             return $this->properties[$name] = $this->model->getRelation($name);
         }
-        if ($this->model->relationLoaded($camelName)) {
-            return $this->properties[$name] = $this->model->getRelation($camelName);
+
+        if (!$this->model->isRelation($name)) {
+            try {
+                $propertyName = $this->model::$snakeAttributes ? Str::snake($name) : $name;
+                return $this->properties[$name] = $this->model->getAttribute($propertyName);
+            } catch (MissingAttributeException) {
+                // Fallback if missing Attribute
+            }
         }
 
         return $this->properties[$name] = UnknownProperty::create();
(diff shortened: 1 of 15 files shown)


Tofandel:
Ended up giving up on the idea of using strict mode, because it has a lot of conditions to throw and a callback that can change the behavior so this wouldn't have been a reliable method

src/Normalizers/Normalized/NormalizedModel.php
```diff
@@ -68,16 +73,13 @@ protected function hasModelAttribute(string $name): bool
             return $this->model->hasAttribute($name);
         }
 
-        // TODO: to use that one once we stop supporting Laravel 10
-
+        // TODO: remove this once we stop supporting Laravel 10
         if (! isset($this->attributesProperty)) {
             $this->attributesProperty = new ReflectionProperty($this->model, 'attributes');
-            $this->attributesProperty->setAccessible(true);
```
Tofandel, line 75: This line does nothing because: https://www.php.net/manual/en/reflectionproperty.setaccessible.php

> Note: As of PHP 8.1.0, calling this method has no effect; all properties are accessible by default.

And we require php 8.1 in comoser.json

src/Normalizers/Normalized/NormalizedModel.php
```diff
@@ -38,10 +39,6 @@ public function getProperty(string $name, DataProperty $dataProperty): mixed
 
     protected function fetchNewProperty(string $name, DataProperty $dataProperty): mixed
     {
-        if ($this->hasModelAttribute($name)) {
-            return $this->properties[$name] = $this->model->getAttributeValue($name);
-        }
-
         $camelName = Str::camel($name);
```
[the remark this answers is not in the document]
Tofandel: Here is the commit for my proposal on how it should be handled in the next major https://github.com/spatie/laravel-data/commit/380942bd1ce14ec0a8040530793054d88085b9c2

src/Normalizers/Normalized/NormalizedModel.php
```diff
@@ -38,10 +39,6 @@ public function getProperty(string $name, DataProperty $dataProperty): mixed
 
     protected function fetchNewProperty(string $name, DataProperty $dataProperty): mixed
     {
-        if ($this->hasModelAttribute($name)) {
-            return $this->properties[$name] = $this->model->getAttributeValue($name);
-        }
-
```
rubenvanassche, line 44: Since we remove this piece of code the whole method on the bottom won't be called anymore so can it be removed?

src/Normalizers/Normalized/NormalizedModel.php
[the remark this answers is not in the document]
rubenvanassche: Yeah I get the idea, I'm also always confused working in this code. But atm we don't have a system in place where we can have different mapping systems for input data. So by setting a mapper on a data object for eloquent models to be snake case all the data you parse from the frontend should also be snake case (which is probably the best idea). But I think we still want people to have an option the manually define another input mapping type for frontend data.

There are complicated solutions to solve this by adding special attributes for this for example, config flags, ... but for now the best solution to me looks like keeping it as is.

src/Normalizers/Normalized/NormalizedModel.php
```diff
@@ -59,6 +56,14 @@ protected function fetchNewProperty(string $name, DataProperty $dataProperty): m
...
         }
 
+        if (! $this->model->isRelation($name) && ! $this->model->isRelation($camelName)) {
+            try {
+                return $this->properties[$name] = $this->model->getAttribute($name);
+            } catch (MissingAttributeException) {
+                // Fallback if missing Attribute
+            }
+        }
+
```
rubenvanassche, line 66: Not sure if we want to do this:

1) Probably 90% of the properties on a model will be regular properties without special logic, now we're checking for every property the relations, attributes, ... this is going to affect performance quite a lot

2) I hope everyone has a minimal amount of missing attributes, but always waiting on exceptions makes code slow and is something to avoid.

rubenvanassche (commented):
In its current state I'm not sure if this is something we want to merge just to solve problems with another package. It introduces some extra performance penalties which we probably want to avoid.

Since normalizers can be overwritten on a data specific object and in config maybe the easiest solution is to add your own normalizer into your project.

If you find a way where we can have less performance issues, please let me know!

src/Normalizers/Normalized/NormalizedModel.php
Tofandel: Well the call to `getAttribute` was done prior to 4.9.0 (which broke this) and has never been a bottleneck

`isRelation` is not an expensive call. It does basically just 1 `isset($mutators[$key])` and one `method_exists($this, $key)` 

The exception catching is there as a bugfix. If you don't enable laravel strict mode it will never happen. But if you do, and try to access a property that doesn't exist on the model then it will throw an exception rather than do the `UnknownProperty` logic of the package which we don't want

Tofandel:
> In its current state I'm not sure if this is something we want to merge just to solve problems with another package. It introduces some extra performance penalties which we probably want to avoid.

Well this is a problem that has been introduced very recently and forces me to pin the package as it breaks everywhere I have it installed

This still fixes the issue while being still way more performant than before https://github.com/spatie/laravel-data/commit/d017301fc37bf61750155847176185571731ae5c

src/Normalizers/Normalized/NormalizedModel.php
rubenvanassche: I'm more worried about the `$dataProperty->attributes->contains` I think we can skip that one by adding a new property to DataProperty isLoadingRelation that way we don't have to check it all the time. Once we have a cached version of that you might be right that the performance wouldn't be affected too much.

Then take a look at the comment on line 41.

If that's done this can be merged.

Tofandel:
I looked into caching all PropertyAttributes, so it should become O(n) instead of O(9n) as multiple contains are made throughout the code and we can just cache it in a single loop, let me know what you think

I'll look into benchmarking before and after to see if it really provides an optimisation as the number of items should still be fairly small in all cases because there is I would say on average 0.3 attributes per property

AJenbo:
> Well this is a problem that has been introduced recently and forces me to pin the package as it breaks everywhere I have it installed

This is true for me as well ... after this PR got merged 😂 Hope you can manage to solve it :)

Issue #882, opened 2024-10-10
Parameters check bug


nsvetozarevic:
**✏️ Describe the bug**
I encountered a rare edge case, which took me quite some time to debug, so I thought it might be useful for every one.
If you need to do any kind of operation in the class's constructor, and something throws `ArgumentCountError` exception, the package will show the exception message that does not point to the real problem.

**↪️ To Reproduce**

```php

it('bug-test', function () {
    class ExampleTestData extends Data
    {
        public function __construct(
            public int $id,
            public string $name,
        ) {
            // Method that throws the `ArgumentCountError` exception
            array_key_exists('test');
        }
    }

    $test = ExampleTestData::from(['id' => 1, 'name' => 'test']);
});
```

**✅ Expected behavior**
So the error message was something like: ```Could not create ExampleTestData: the constructor requires 2 parameters, 2 given. Parameters given: id, name. Parameters missing: .```

If you guys are interested, I could try to submit a PR to fix it?
I made it to work with the changes to `DataFromArrayResolver`. Instead of checking for `ArgumentCountError` exception, I added a method to check if all the parameters are supplied. That way, If an `ArgumentCountError` is thrown it would point to the direct issue.

```
    if (! $this->allParametersSupplied($dataClass, array_keys($parameters))) {
        throw CannotCreateData::constructorMissingParameters(
            $dataClass,
            $parameters
        );
     }

    protected function allParametersSupplied(DataClass $dataClass, array $parameters): bool
    {
        $dataClassConstructorParameters = $dataClass->constructorMethod->parameters->pluck('name')->toArray();
        
        return count(array_diff($dataClassConstructorParameters, $parameters)) === 0;
    }
```

`CannotCreateData::constructorMissingParameters()` would have to be updated as well as there would be no `$previous` Throwable.

**🖥️ Versions**

Laravel: 11.27.2
Laravel Data: 4.10.1
PHP: 8.3.3

rubenvanassche:
Hmm good catch, you're certainly welcome to send in a PR!

PR #884, opened 2024-10-10
Point to the right problem on ArgumentCountError exception

nsvetozarevic:
Fixes https://github.com/spatie/laravel-data/issues/882

### DIFF
--- a/src/Exceptions/CannotCreateData.php
+++ b/src/Exceptions/CannotCreateData.php
@@ -24,7 +24,6 @@ class CannotCreateData extends Exception
     public static function constructorMissingParameters(
         DataClass $dataClass,
         array $parameters,
-        Throwable $previous,
     ): self {
         $parameters = collect($parameters);
 
@@ -42,6 +41,6 @@ class CannotCreateData extends Exception
                 ->map(fn (DataProperty|DataParameter $parameter) => $parameter->name)
                 ->join(', ')}.";
 
-        return new self($message, previous: $previous);
+        return new self($message);
     }
 }
--- a/src/Exceptions/CannotCreateData.php
+++ b/src/Exceptions/CannotCreateData.php
@@ -24,6 +24,7 @@ class CannotCreateData extends Exception
     public static function constructorMissingParameters(
         DataClass $dataClass,
         array $parameters,
+        Throwable $previous,
     ): self {
         $parameters = collect($parameters);
 
@@ -41,6 +42,6 @@ class CannotCreateData extends Exception
                 ->map(fn (DataProperty|DataParameter $parameter) => $parameter->name)
                 ->join(', ')}.";
 
-        return new self($message);
+        return new self($message, previous: $previous);
     }
 }
--- a/src/Resolvers/DataFromArrayResolver.php
+++ b/src/Resolvers/DataFromArrayResolver.php
@@ -91,14 +91,23 @@ class DataFromArrayResolver
             }
         }
 
-        try {
-            return new $dataClass->name(...$parameters);
-        } catch (ArgumentCountError $error) {
+        if ($this->isAnyParameterMissing($dataClass, array_keys($parameters))) {
             throw CannotCreateData::constructorMissingParameters(
                 $dataClass,
                 $parameters,
-                $error
             );
         }
+
+        return new $dataClass->name(...$parameters);
+    }
+
+    protected function isAnyParameterMissing(DataClass $dataClass, array $parameters): bool
+    {
+        return $dataClass
+            ->constructorMethod
+            ->parameters
+            ->pluck('name')
+            ->diff($parameters)
+            ->isNotEmpty();
     }
 }
--- a/src/Resolvers/DataFromArrayResolver.php
+++ b/src/Resolvers/DataFromArrayResolver.php
@@ -93,14 +93,19 @@ class DataFromArrayResolver
             }
         }
 
-        if ($this->isAnyParameterMissing($dataClass, array_keys($parameters))) {
-            throw CannotCreateData::constructorMissingParameters(
-                $dataClass,
-                $parameters,
-            );
+        try {
+            return new $dataClass->name(...$parameters);
+        } catch (ArgumentCountError $error) {
+            if ($this->isAnyParameterMissing($dataClass, array_keys($parameters))) {
+                throw CannotCreateData::constructorMissingParameters(
+                    $dataClass,
+                    $parameters,
+                    $error
+                );
+            } else {
+                throw $error;
+            }
         }
-
-        return new $dataClass->name(...$parameters);
     }
 
     protected function isAnyParameterMissing(DataClass $dataClass, array $parameters): bool
--- a/src/Resolvers/DataFromArrayResolver.php
+++ b/src/Resolvers/DataFromArrayResolver.php
@@ -104,8 +104,6 @@ class DataFromArrayResolver
                 throw $error;
             }
         }
-
-        return new $dataClass->name(...$parameters);
     }
 
     protected function isExceptionCausedByDataClass(ArgumentCountError $error, DataClass $dataClass): bool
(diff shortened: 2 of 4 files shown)


Tofandel:
You could also just check the trace of the exception and run some reflection in the catch to decide wether to throw the original error or not https://onlinephp.io/

src/Resolvers/DataFromArrayResolver.php
```diff
@@ -91,14 +91,23 @@ protected function createData(
             }
         }
 
-        try {
-            return new $dataClass->name(...$parameters);
-        } catch (ArgumentCountError $error) {
+        if ($this->isAnyParameterMissing($dataClass, array_keys($parameters))) {
```
Tofandel, line 94:
```suggestion
        try {
            return new $dataClass->name(...$parameters);
        } catch (ArgumentCountError $error) {
            $trace = $error->getTrace()[0];
            if ($trace['function'] === '__construct' && $trace['class'] === (new ReflectionMethod($dataClass->name, '__construct'))->getDeclaringClass()->getName()) {
                throw CannotCreateData::constructorMissingParameters(
                    $dataClass,
                    $parameters,
                    $error
                );
            } else {
                throw $error;
            }
        }
```

nsvetozarevic:
@Tofandel Optional parameters should work actually. Hence all optinal parameters have a default value and they are present in `$parameters` array. I added an optional parameter to the test case.

As for variadic parameters, I really couldn't figure out how would it create problems? Can you please show me an example?

On the other hand if you're really against this approach, I could check the trace and rethrow the exception.

Tofandel:
> have a default value and they are present in $parameters array

That is only the case for non promoted parameters, promoted parameters do not appear in the array, according to the code, this would cause issues

https://onlinephp.io?s=nVJbS8MwFH5eof_hDMZIYGPq46oTmb7KEPFljJJlpzbaJSVJq0X2301CXetlL4Y-lO92vpP28rrMyzjiBTMGHtFY-IijQalFzSyCkBZG1qGJB6ttIThkleRWKAlpypU0VlfcEkcPjmK4grOJR37FXDjq3FM0jBmMbC7MdNGajpMOceSeOJrN4E5a1NCoSgNXO4QcNU4A5YtqhnE04s5GJL7BA2YFhl5Lvwrxq8znYS1Kp4tntMuvskoTmnhvB6yYZnt0k4yvwYO-w4LcGUqtStRWoFetNz6j7BsDlCmNjOdATuQzA50rXAO4IzIgTGvWpK_YpPgujDWk04VC9-6d0An0etBjgD-9Nus_vRu_Xec-IUq6RLeCFbLCFjp0ZYfQNwuz0mqvLO4IhfH4B3eLGasKW7OiwpuaiYJtCzfoP92_EW3uk88N36jtGP6dmul0V-3L3i0amnwC&v=8.3.8%2C8.2.20%2C8.1.29

src/Resolvers/DataFromArrayResolver.php
```diff
@@ -91,14 +91,23 @@ protected function createData(
...
+
+        return new $dataClass->name(...$parameters);
+    }
+
+    protected function isAnyParameterMissing(DataClass $dataClass, array $parameters): bool
+    {
+        return $dataClass
+            ->constructorMethod
+            ->parameters
+            ->pluck('name')
```
Tofandel, line 109:
```suggestion
            ->filter(fn ($p) => !$parameter->hasDefaultValue)
            ->pluck('name')
```

nsvetozarevic:
Wow, needed some time to wrap my head around this :) Sorry for the late reply.

> promoted parameters do not appear in the array

I can now see the confusion. They do appear. They are passed to `createData` function as part of the `properties` array.  (`DataFromArrayResolver` class) .

> can you check your new test case?

I did run my test case, and it passed. I think that the optional parameters are covered. In the example you provided, promoted parameter was private, that's why it didn't show up in `$parameters`, but it was also not present in the `$dataClass->constructorMethod->parameters`... 


Variadic parameter on the other hand is a problem. As you stated, they are always optional, and if not supplied, they are not present in the `parameters`.  I can't find a way to filter them out. Optional parameters have the default value, so it would be easy to filter them out, but variadic do not.

> ->filter(fn (\Spatie\LaravelData\Support\DataParameter $p) => !$parameter->hasDefaultValue)

Wouldn't this filter out required parameters as well, which do not have a default value?

Having all this in mind I think I'll have to switch back to checking the stack trace. Great catch btw  👍

nsvetozarevic:
So I switched back to checking the trace, but it's not that straightforward either.
```
Trace of the Eeception caused by not passing right parameters to the data class
array:6 [
  "file" => "/Users/nikolasvetozarevic/Projects/laravel-data-parameters-check/src/Resolvers/DataFromArrayResolver.php"
  "line" => 95
  "function" => "__construct"
  "class" => "Spatie\LaravelData\Tests\Fakes\DataWithArgumentCountErrorException"
  "type" => "->"
  "args" => array:2 [
    0 => "string"
    1 => null
  ]
] 

Trace of the Exception caused by something else in the constructor
array:6 [
  "file" => "/Users/nikolasvetozarevic/Projects/laravel-data-parameters-check/src/Resolvers/DataFromArrayResolver.php"
  "line" => 95
  "function" => "__construct"
  "class" => "Spatie\LaravelData\Tests\Fakes\DataWithArgumentCountErrorException"
  "type" => "->"
  "args" => array:5 [
    0 => "string"
    1 => null
    2 => "default"
    3 => "optional"
    4 => "test"
  ]
] 

```
As you can see, they are basically identical. So the only Idea I have is to check the exceptions message. But that feels kind of hacky. I'll push those changes, please tell me what do you think.

Tofandel:
I think it does not need to be 100% foolproof to the exact same exception thrown in the body of the same function, because in real life it wil happen somewhere else or on another function, because in the end it's just exception wrapping meant for debugging.

I do wonder if maybe the laravel utils just don't provide a way out of the box to check the parameters of a class (but it might have the side effects of adding dependency injection to the constructor, which is not necessarily a bad thing)

src/Resolvers/DataFromArrayResolver.php
Tofandel: > Wouldn't this filter out required parameters as well, which do not have a default value?

It would keep only parameters without default values aka required ones, though there is one obscure area, is if you have a default value before a required param (which normally emits a warning)

Which is why filtering on is optional would provide the perfect solution, but it would need to be added to the DataParameter class and factory..
Or a new Reflection needs to be made on the spot

nsvetozarevic:
@Tofandel I reverted the code back to check the parameters and added the filtering you suggested. Is this what you meant?

nsvetozarevic:
@Tofandel just checking if the PR is still alive :)

src/Exceptions/CannotCreateData.php
```diff
@@ -42,6 +41,6 @@ public static function constructorMissingParameters(
                 ->map(fn (DataProperty|DataParameter $parameter) => $parameter->name)
                 ->join(', ')}.";
 
-        return new self($message, previous: $previous);
```
Tofandel, line 45: Looks good to me, except this deletion

nsvetozarevic: Previously `ArgumentCountError` was thrown, but now hence we do the checks based on the data class properties, I had to remove that because there is no previous error.

Tofandel: I must have been squinting and saw my suggested change as the code. Have a look at the suggestion, basically, keep the try catch but do the check only if it throws, like this you let php handle the argument validation

nsvetozarevic:
@Tofandel Sorry for not getting back to you sooner. I applied what you suggested. Tests are passing for me locally.

Tofandel (approved):

rubenvanassche:
Thanks!

PR #891, opened 2024-10-19
Add ability to set responseStatusCode on Response objects

cosmastech:
## Goal
Provide the ability to set a specific status code on responses.

## Background
I'm using the package for error response messages (rather than throwing exceptions). For instance, we have a custom NotAuthorizedResponse that will have a payload like this:
```json
{
    "code": "not_authorized",
    "uuid": "abcd-e4gab2-ffff-a324-0001",
    "message": "You are not authorized to view this post"
}
```

And an object like this:

```php
class NotAuthorizedResponse extends Response
{
    public ErrorResponseEnum $code = ErrorResponseEnum::NOT_AUTHORIZED;

    public function __construct(
        public ?string $uuid,
        public string $message = "We could not find this resource"
    ) {}
}
```

We can get away with overriding the `calculateResponseStatus` method in a `BaseResponse` class, but I thought that this functionality would be helpful for others as well.

## Usage
```php
public function show(Request $request): PostResponse|NotAuthorizedResponse
{
    if (! $request->user()->isAdmin()) {
        return (new NotAuthorizedResponse($request->uuid, "You are not authorized to view this post"))
            ->setResponseStatusCode(403);
    }
   // ... happy path
}
```

### DIFF
--- a/src/Concerns/ResponsableData.php
+++ b/src/Concerns/ResponsableData.php
@@ -12,6 +12,8 @@ use Spatie\LaravelData\Support\Wrapping\WrapExecutionType;
 
 trait ResponsableData
 {
+    protected int $responseStatusCode;
+
     public function toResponse($request)
     {
         $contextFactory = TransformationContextFactory::create()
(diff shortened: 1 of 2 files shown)


rubenvanassche:
I think we want to keep the amount of properties within people's data objects to the bare minimum so adding another one just for setting a response code seems a lot since it will be included in every data object.

In such a case, why not create a trait which overwrites `calculateResponseStatus` with a 403?

cosmastech:
> I think we want to keep the amount of properties within people's data objects to the bare minimum so adding another one just for setting a response code seems a lot since it will be included in every data object.

No worries, I can understand not wanting the package to expand for little edge cases. 👍 This package is great and my team (frontend and backend) really appreciates it!
 
> In such a case, why not create a trait which overwrites `calculateResponseStatus` with a 403?

That's exactly what we have done already.

PR #896, opened 2024-10-29
Implement enum json serialization

dont-know-php:
Hi team, this is a small PR fixing enum serialization.

## Prerequisites

[zend.exception_ignore_args](https://www.php.net/manual/en/ini.core.php) must be disabled.

## Description

PHP doesn't support Pure Enum json serialization: https://www.php.net/manual/en/language.enumerations.serialization.php

> If a Pure Enum is serialized to JSON, an error will be thrown.

We will get an serialization error, for example, if we want to write json serialized error trace with arguments into log, because some package classes/methods use enum in arguments.

There are two possible solutions:
1. Using Backed Enum
2. Implement `JsonSerializable` interface. I chose this option because the package uses Pure Enums.

### DIFF
--- a/src/Enums/CustomCreationMethodType.php
+++ b/src/Enums/CustomCreationMethodType.php
@@ -2,9 +2,9 @@
 
 namespace Spatie\LaravelData\Enums;
 
-enum CustomCreationMethodType
+enum CustomCreationMethodType: string
 {
-    case None;
-    case Object;
-    case Collection;
+    case None = 'None';
+    case Object = 'Object';
+    case Collection = 'Collection';
 }
--- a/src/Enums/DataCollectableType.php
+++ b/src/Enums/DataCollectableType.php
@@ -2,11 +2,11 @@
 
 namespace Spatie\LaravelData\Enums;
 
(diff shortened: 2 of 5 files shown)


rubenvanassche:
Hi @dont-know-php,

I'm not sure when it happens that these structures would be converted into JSON. I'll guess you're catching stacktraces with arguments and somehow try to JSON encode them to send them to an external service. But I think it then would be easier just to make them string backed enums instead of building an own serialization mechanism.

dont-know-php:
hi @rubenvanassche ,

> I'll guess you're catching stacktraces with arguments and somehow try to JSON encode them to send them to an external service

You are absolutely right, by default stackraces with arguments is disabled using [zend.exception_ignore_args](https://www.php.net/manual/en/ini.core.php) ini configuration directive. But we use detailed tracing with arguments in our Laravel projects.

> But I think it then would be easier just to make them string backed enums instead of building an own serialization mechanism.

Agree, I'have pushed new changes with Backed enums implementation. Please, advise.

rubenvanassche:
Thanks!

PR #904, opened 2024-11-28
Use comments instead of docblocks in configuration file.

edwinvdpol:
Hi there 👋 

This PR replaces the PHP docblocks with comments in the configuration file.

I also updated the location of the Larastan package as mentioned [here](https://github.com/larastan/larastan/blob/3.x/UPGRADE.md#upgrading-to-270-from-265).

Thanks!


rubenvanassche:
Thanks!

PR #908, opened 2024-12-04
Casting DateTimeInterface: Truncate nanoseconds to microseconds (first 6 digits) / with Tests

yob-yob:
So I'm basically just making this pull request to fix a very specific issue I'm having.

a Third Party API i'm using is returning dates in ISO 8601 Format with Nanoseconds precision, but sadly PHP does not support nanoseconds (they only support up-to microseconds)

Example:
A provided datetime 2024-12-03T10:33:18.924711041Z is in ISO 8601 format with a nanosecond precision and a timezone indicator (Z for UTC). However, PHP's DateTime and the provided formats DATE_ATOM or Y-m-d\TH:i:s.u\Z do not handle nanoseconds (fractions of a second) by default, which is causing the error.

### DIFF
--- a/src/Casts/DateTimeInterfaceCast.php
+++ b/src/Casts/DateTimeInterfaceCast.php
@@ -38,6 +38,11 @@ class DateTimeInterfaceCast implements Cast, IterableItemCast
             return Uncastable::create();
         }
 
+        // Truncate nanoseconds to microseconds (first 6 digits)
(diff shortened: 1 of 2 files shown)


yob-yob:
also... when I made a fresh new branch from `main` and ran `pest`... 

There was some errors already on the test suite, am I required to fix those test to make this pull-request valid?

rubenvanassche:
Thanks!

PR #910, opened 2024-12-05
Use container to call `Data::authorize()` to allow for dependencies

cosmastech:
I want to build a request object that leverages Laravel's new contextual attributes. For instance:

```php
class CreatePodcast extends Data
{
    public function __construct(
        public string $name,
        public bool $premium
    ) {
    }

    public static function authorize(
        #[CurrentUser] User $user,
        #[RouteParameter('org')] Organization $org
    ): bool
    {
        return $user->can('createPodcastFor', $org);
    }
}
```

While it's possible to add a custom DataPipe to `pipelines()`, this seems like it would be a pretty common use-case for folks.


rubenvanassche:
Thanks!

PR #914, opened 2024-12-24
Use isAbstractClassCast in DataCollectionEloquentCast for easy override

bentleyo:
I'm using abstract data classes in a way that's different to the eloquent-cast functionality of `laravel-data` (related to the problem I described in my old example PR https://github.com/spatie/laravel-data/pull/304).

The approach I have now is different / cleaner as I move from V2 to V3, but I need to be able to disable the abstract class functionality in `DataCollectionEloquentCast` and I'd prefer not to duplicate the code in the casts if possible!

### DIFF
--- a/src/Support/EloquentCasts/DataCollectionEloquentCast.php
+++ b/src/Support/EloquentCasts/DataCollectionEloquentCast.php
@@ -108,6 +108,7 @@ class DataCollectionEloquentCast implements CastsAttributes
     protected function isAbstractClassCast(): bool
     {
         $dataClass = $this->dataConfig->getDataClass($this->dataClass);
+
         return $dataClass->isAbstract && $dataClass->transformable;
     }
 }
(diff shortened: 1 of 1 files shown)


bentleyo:
If you're curious, what I'm doing is using properties inside the DTO to describe which concrete class / variant of the abstract class to use (as per https://github.com/spatie/laravel-data/pull/304). I've got support for creation from request payloads (with validation), singular data classes, data collections and nesting.

An example use case is having `Vehicle` as an abstract data class with `type` (to determine which concrete to use) and a common `passengers` property. Each concrete can then add additional properties e.g.:
```
"vehicles": [
    { "type": "horse", "max_passengers": 1, "name": "Bill" },
    { "type": "car", "max_passengers": 5, "transmission": "manual" },
]
```

It's something that's been super useful for us and a feature I think would be really great to have built-in. If you're curious about how I'm solving that problem I'm happy to share what I've got (outside this PR of course) 😄

bentleyo:
**Update:** I've created a PR for this here: https://github.com/spatie/laravel-data/pull/921

rubenvanassche:
Gonna close this one in favor of #921

Issue #913, opened 2024-12-20
Illuminate\Database\Eloquent\Casts\ArrayObject not transforming into array


qbaniewiadomski:
**✏️ Describe the bug**
In User model I have cast database column `roles` (which contains enum values in array format e.g. `["ADMIN","EMPLOYEE"]`) to ArrayObject with Enums.
```
Illuminate\Database\Eloquent\Casts\ArrayObject {#681 // app/Http/Controllers/UserController.php:118
  storage: array:1 [
    0 => 
App\Enum
\
UserRole {#687
      +name: "ADMIN"
      +value: "ADMIN"
    }
  ]
  flag::STD_PROP_LIST: false
  flag::ARRAY_AS_PROPS: false
  iteratorClass: "ArrayIterator"
}
```

When trying to return it in DTO as array, I receive:
```
App\\DTO\\Response\\User\\UserDto::__construct(): Argument #8 ($roles) must be of type array, Illuminate\\Database\\Eloquent\\Casts\\ArrayObject given, called in /app/vendor/spatie/laravel-data/src/Resolvers/DataFromArrayResolver.php on line 95
```

It worked fine in version `4.6.0`
Currently I did workaround:
```php
final class UserDto extends Data
{
    /** @var array<UserRole> $roles */
    #[Computed]
    public readonly array $roles;

    public function __construct(
        ArrayObject $roles,
    ) {
        $this->roles       = $roles->getArrayCopy();
    }
}
```

**↪️ To Reproduce**

User Model:
```php
class User extends Authenticatable
{
[...]
    protected function casts(): array
    {
        return [
            'roles' => AsEnumArrayObject::class . ':' . UserRole::class,
        ];
    }
[...]
}
```
UserDto Class:
```php
final class UserPreviewDto extends Data
{
    [...]
    /** @param array<UserRole> $roles */
    public function __construct(
        [...]
        public readonly array $roles = [],
    ) {
        [...]
    }
}
```
Than: `UserDto::from($user)`

**✅ Expected behavior**
`Illuminate\Database\Eloquent\Casts\ArrayObject` should be casted to array, e.g.:
```
["ADMIN"]
```

**🖥️ Versions**

Laravel: 11.33.2
Laravel Data: 4.11.1
PHP: 8.3.14

rust17:
[...]

PR #916, opened 2024-12-24
Casts an array of enum values to array

rust17:
see https://github.com/spatie/laravel-data/issues/913


rubenvanassche:
Thanks, closed due to comment in issue.

PR #917, opened 2024-12-25
Failing annotation tests

bentleyo:
Tests were failing for me locally and on some branch builds. E.g. for the same commit:
- This one passed: https://github.com/spatie/laravel-data/actions/runs/12477953676/job/34824836963
- This one failed: https://github.com/spatie/laravel-data/actions/runs/12477953676/job/34824836512

I believe this is happening because of an update to `phpdocumentor/reflection-docblock` that added support for more specific docblock types: https://github.com/phpDocumentor/ReflectionDocBlock/releases/tag/5.5.0

This means that the returned tag classes for the `@extends` and `@template` are new specific classes instead of the "Generic" class they previously were.

Unfortunately this dependency is loaded indirectly via  `phpdocumentor/reflection` and their requirement for `phpdocumentor/reflection-docblock` is just `^5`. So I've added `^5.5` to ensure we have the required classes and to make the check more consistent.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
         "php": "^8.1",
         "illuminate/contracts": "^10.0|^11.0",
         "phpdocumentor/reflection": "^6.0",
+        "phpdocumentor/reflection-docblock": "^5.5",
         "spatie/laravel-package-tools": "^1.9.0",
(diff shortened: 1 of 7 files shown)


freekmurze:
This was fixed by another PR

PR #918, opened 2024-12-26
Fix GitHub action fail

rust17:
1. Fix phpstan issue
2. Make it compatible with `phpdocumentor/reflection-docblock` 5.6.1. Several changes were introduced in `phpdocumentor/reflection-docblock` version [5.4.0](https://github.com/phpDocumentor/ReflectionDocBlock/releases/tag/5.4.0) and later. These changes may affect existing usage. The core functionality remains the same, as evidenced by the unchanged test cases. We're working to ensure a smooth transition for everyone.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -121,7 +121,7 @@ parameters:
 			path: src/Support/DataConfig.php
 
 		-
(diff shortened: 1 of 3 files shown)


freekmurze:
Thanks!

PR #920, opened 2024-12-30
General code health improvements

xHeaven:


xHeaven:
How do we feel about being _very_ explicit with `if` conditions? We could merge things like this into a single if statement instead of branching out a bunch of them:
```php
if (!is_object($value)) {
    return true;
}

if ($property->type->kind->isDataCollectable()) {
    return true; // Transform everything to data objects
}

// ---

$class = new ReflectionClass($className);

if (empty($class->getDocComment())) {
    return self::$cache[$className] = null;
}

if (! $this->isIterable($class)) {
    return self::$cache[$className] = null;
}

$type = $this->getCollectionReturnType($class);

if ($type === null || $type['valueType'] === null) {
    return self::$cache[$className] = null;
}

// ---

if ($property->type->isNullable || $property->type->isOptional) {
    return false;
}

if ($property->type->kind->isDataCollectable() && $rules->hasType(Present::class)) {
    return false;
}

if ($rules->hasType(BooleanType::class)) {
    return false;
}

if ($rules->hasType(Nullable::class)) {
    return false;
}

if ($rules->hasType(RequiringRule::class)) {
    return false;
}
```

xHeaven:
I'm not entirely sure about [`e1395b1` (#920)](https://github.com/spatie/laravel-data/pull/920/commits/e1395b1d394930e0ae1497a20eb8c987f6d82899), is this the intended behavior? There are no tests for `false` expectations.

xHeaven:
For the failed test, I can either do this:
```php
$collection = array_map(
    fn ($value) => $this->transformationClosure($nestedContext)($value),
    is_array($items) ? $items : $items->toArray()
);
```

Or I can revert the entire change to a foreach loop. Which one is preferred?

src/Concerns/AppendableData.php
```diff
@@ -24,13 +24,10 @@ public function getAdditionalData(): array
     {
         $additional = $this->with();
 
-        $computedAdditional = [];
-
-        foreach ($additional as $name => $value) {
-            $computedAdditional[$name] = $value instanceof Closure
-                ? ($value)($this)
-                : $value;
-        }
```
rubenvanassche, line 33: Foreach's are a lot more performant than array_map let's keep this as is

src/Resolvers/DataValidationMessagesAndAttributesResolver.php
```diff
@@ -63,13 +64,14 @@ public function execute(
...
-
-                continue;
+                $messages[] = $collected['messages'];
+                $attributes[] = $collected['attributes'];
             }
         }
 
+        $messages = array_merge(...$messages);
+        $attributes = array_merge(...$attributes);
+
```
rubenvanassche, line 74: Good one, this is probably a lot more performant

src/Resolvers/TransformedDataCollectableResolver.php
```diff
@@ -75,11 +75,10 @@ protected function transformItems(
...
     ): array {
-        $collection = [];
-
-        foreach ($items as $key => $value) {
-            $collection[$key] = $this->transformationClosure($nestedContext)($value);
-        }
+        $collection = array_map(
+            fn ($value) => $this->transformationClosure($nestedContext)($value),
+            is_array($items) ? $items : $items->all()
+        );
```
rubenvanassche, line 81: Same as above, we want the performance of foreach

rubenvanassche:
As for your comments:

1) Let's keep the if's as is, yeah the conditions are quite large but in my opinion it's easier to see which path's exist

2) Yeah that was apparently a problem, good catch!

3) Foreach is faster so foreach 😄 

If the changes are made, this can be merged!

rubenvanassche (changes requested):

src/Concerns/AppendableData.php
xHeaven: > Foreach's are a lot more performant than array_map let's keep this as is

Do you, by any chance, have any benchmarks on this? I ran a few benchmarks between the two, and I couldn't find any measurable difference in performance. Also, I feel like `array_map` reads better. Update me on this if you have the time, if anything, I'll revert it back to `foreach` if you insist. Thanks!

```php
use Illuminate\Support\Benchmark;

$additional = [];
for ($i = 0; $i < 10000; $i++) {
    $additional["key_$i"] = $i % 2 === 0 
        ? fn($obj) => "computedValue_$i"
        : "value_$i";
}

$results = Benchmark::measure([
    'foreach' => function () use ($additional) {
        $computedAdditional = [];
        foreach ($additional as $name => $value) {
            $computedAdditional[$name] = $value instanceof Closure
                ? ($value)(new stdClass)
                : $value;
        }
        return $computedAdditional;
    },
    'array_map' => function () use ($additional) {
        $computedAdditional = array_map(
            fn($value) => $value instanceof Closure ? ($value)(new stdClass) : $value,
            $additional
        );
        return $computedAdditional;
    },
], iterations: 10000);

// Results
//[
//    "foreach" => 1.1958281731,
//    "array_map" => 1.1976062882,
//]

src/Resolvers/TransformedDataCollectableResolver.php
xHeaven: > Same as above, we want the performance of foreach

See https://github.com/spatie/laravel-data/pull/920#discussion_r1921085756.

src/Concerns/AppendableData.php
xHeaven: Fair enough, I'll revert those changes in a bit.

xHeaven: @rubenvanassche reverted the `array_map`s, let me know if you see any other issues, if not, feel free to merge.

rubenvanassche: Thanks!

PR #921, opened 2025-01-02
Feature: property-morphable abstract data

bentleyo:
This PR adds a new feature where abstract data classes can dynamically choose which concrete variant of the class to use based on the payload. This feature works with validation, collections and nested payloads.

While the existing abstract class functionality is good, I find it to be limiting for our use case as it's only available when casting. I've maintained this existing functionality without any breaking changes.

An example use case is having `Vehicle` as an abstract data class with `type` (to determine which concrete to use) and a common `max_passengers` property used by every class that extends `Vehicle`. Each concrete can then add its own additional properties e.g.:

```json
"vehicles": [
    { "type": "horse", "max_passengers": 1, "name": "Bill" },
    { "type": "car", "max_passengers": 5, "transmission": "manual" },
]
```

This is a feature that's been super useful for us (we have it working on v3 using a trait, but that's no longer possible) and it's a feature that I think would be really great to have built-in.

To use the feature you implement the new `PropertyMorphableData` contract and associated static `morph` method. This method is supplied with an array of properties and should return a class string.

E.g.

```php
abstract class VehicleData extends Data implements PropertyMorphableData
{
    public function __construct(
        #[In('car', 'horse')]
        public string $type,
    ) {
    }

    {
        return match ($properties['type'] ?? null) {
            'car' => CarData::class,
            'horse' => HorseData::class,
            default => null,
        };
    }
}
```


bentleyo:
The reason the return value for `morph` is nullable is because we need to know which type of class we're building when generating validation rules. However, at this point not all required properties may be available or supplied.

Sagmedjo:
Really looking forward to this feature. Is it going to be in an upcoming release?

src/Resolvers/DataFromSomethingResolver.php
```diff
@@ -39,10 +40,9 @@ public function execute(
         $payloadCount = count($payloads);
 
         if ($payloadCount === 0 || $payloadCount === 1) {
-            return $this->dataFromArrayResolver->execute(
-                $class,
-                $pipeline->execute($payloads[0] ?? [], $creationContext)
-            );
+            $properties = $pipeline->execute($payloads[0] ?? [], $creationContext);
+
+            return $this->dataFromArray($class, $creationContext, $payloads, $properties);
```
rubenvanassche, line 45: I'm not completely sure why we first run the pipeline here with the abstract class and then run it again with the inherited class? That first run seems to be completely unnecessary right?

src/Resolvers/DataValidationRulesResolver.php
```diff
@@ -34,6 +35,15 @@ public function execute(
     ): array {
         $dataClass = $this->dataConfig->getDataClass($class);
 
+        if ($this->isPropertyMorphable($dataClass)) {
+            /**
+             * @var class-string<PropertyMorphableData> $class
+             */
+            $payload = $path->isRoot() ? $fullPayload : Arr::get($fullPayload, $path->get(), []);
+            $class = $class::morph($payload) ?? $class;
+            $dataClass = $this->dataConfig->getDataClass($class);
```
rubenvanassche, line 44: Let's rewrite this to:

```suggestion
            $morphedClass = $class::morph(
                $path->isRoot() ? $fullPayload : Arr::get($fullPayload, $path->get(), [])
            );
            
            $dataClass = $this->dataConfig->getDataClass($morphedClass  ?? $class);
```

tests/CreationTest.php
```diff
@@ -53,6 +53,10 @@
 use Spatie\LaravelData\Tests\Fakes\NestedLazyData;
 use Spatie\LaravelData\Tests\Fakes\NestedModelCollectionData;
 use Spatie\LaravelData\Tests\Fakes\NestedModelData;
+use Spatie\LaravelData\Tests\Fakes\PropertyMorphableData\AbstractPropertyMorphableData;
+use Spatie\LaravelData\Tests\Fakes\PropertyMorphableData\NestedPropertyMorphableData;
+use Spatie\LaravelData\Tests\Fakes\PropertyMorphableData\PropertyMorphableDataA;
+use Spatie\LaravelData\Tests\Fakes\PropertyMorphableData\PropertyMorphableDataB;
```
rubenvanassche, line 59: Let's try to inline these classes as much as possible, the Fakes directory is already sooo big with a lot of specific classes. My goal is trying to eliminate  90% of them one day so that the definitions of the classes are closer to the tests and so that we don't have to come up with more class names 😄.

Feel free to use anonymous classes or inline classes (just prefix the names with Test).

If a class is being used a lot it might be a fit to put in Fakes but most of the cases it isn't. Except for classes with class attributes because they aren't parsed when written inline if I'm correct.

rubenvanassche (changes requested):
Hi @bentleyo,

This is a really cool PR! Would love to merge it but have a few question/suggestions. Once that has been worked out I think this one can be merged.

There are some points I couldn't add to the review:

- What would happen if we transform an morphed class? Of course it will just be converted into an array/json/... but since we also have support for including/excluding properties will that still work? Especially if a property exists in one class but not in another, again the same with nested structures. Tbh, I have no idea how the code would react we haven't taught about this when building the partials system. Some tests here would be really useful.
- At the moment there aren't docs on this new feature, which I'm sure a lot of people want to use. Would be cool to also include this!

Would love to merge this one! As (probably) my child will be born next week I will be on parental leave for four weeks so could be I'm not directly approving it 😄

bentleyo:
@rubenvanassche thank you for reviewing! I'll respond over the next couple of days. This is a feature I'm willing to put effort into getting right 😊

How exciting! Congratulations 🎉 Enjoy the next few weeks with your family!

src/Resolvers/DataValidationRulesResolver.php
bentleyo: I've updated this 👍

tests/CreationTest.php
bentleyo: @rubenvanassche I've had a go at inlining the classes. Can you have a look and let me know what you think?

Unfortunately it's a bit difficult to work with the abstract classes anonymously as I can't create an instance of them (e.g. `new abstract class ()` wouldn't work). I also wanted to avoid duplicating all the classes I had in `CreationTest.php` many times (especially as the classes would need unique names), so for now I've created a `describe` section and put the class definitions at the top. Are you happy with that as a solution?

bentleyo:
@rubenvanassche I'm happy to put together some documentation for this feature, but before I do I wanted to point out some potential awkwardness with my current approach and see if you had any thoughts. Sorry for the long message 😅 

---

#### Potential Problem 1
At the moment the code depends on the user adding validation rules on their abstract class to ensure that a concrete class can be determined. For example, in `ValidationTest.php` I have:

```
abstract class TestValidationAbstractPropertyMorphableData extends Data implements PropertyMorphableData
{
    public function __construct(
        #[In('a', 'b')]
        public string $variant,
    ) {
    }

    {
        return match ($properties['variant'] ?? null) {
            'a' => TestValidationPropertyMorphableDataA::class,
            'b' => TestValidationPropertyMorphableDataB::class,
            default => null,
        };
    }
}
```

Without the `In` validation, if you provided `{ "variant": "c" }` as data  it would pass validation, but then fail on creation because it wouldn't be able to determine which concrete class to use.

#### Potential Problem 2
The morph method gets called from two places. On _validation_ when the properties are usually uncast (e.g. from a request)  or from _creation_ when the properties  have been cast (see my other comment https://github.com/spatie/laravel-data/pull/921#discussion_r1931844381). This means that if you're checking a cast property, e.g. an enum, that you'd need to handle the situation it's a string OR an enum instance.

---

Do you have any thoughts on those potential problems?

The reason I implemented it like this is because our particular situation uses two properties to determine the concrete class to use. E.g. `platform_type` and `module_type` which are both enums.

One potential solution to the above problems is to make it so the abstract classes only have one property used to determine the concrete class and have it always be a string e.g. `public string $type;`.

This would allow us to avoid the overhead of casting properties before they go to `morph` on creation and allow us to add a validation error if the property cannot be used to determine a concrete class.

bentleyo:
@rubenvanassche I have created a PR on my repo that resolves the two problems I identified in https://github.com/spatie/laravel-data/pull/921#issuecomment-2618626779 while retaining the ability for us to use multiple properties to determine the morphed class and supporting casts 🎉 

Can you let me know if you would be happy with this approach?
https://github.com/bentleyo/laravel-data/pull/1

bentleyo:
@rubenvanassche hope all is well with your family and that you got to enjoy some quality time with your newborn ☺️

I've merged the PR I linked in my previous comment to simplify things and make this PR easier to get back into reviewing.

Let me know if there's anything you'd like me to do to help get this moving 🙏 

If you didn't see that PR it had this description:

> This PR fixes the problems identified in https://github.com/spatie/laravel-data/pull/921#issuecomment-2618626779.
> 
> I've achieved this by doing the following:
> - Requiring a `#[PropertyForMorph]` attribute to be added to property-morphable properties
>   - By doing this it allows us to add validation messages if the concrete version of the data class cannot be determined
>   - It also allows us to restrict the properties passed into morph and only run it if they are all available
> - Casting properties when validating
>   - This allows us to use more complex types e.g. enums or values with defaults for morphable properties without having to worry if the provided value has been cast already or not

rubenvanassche:
Hi @bentleyo,

Thanks for the changes, started going through your PR this week and tweaking some things here and there, will continue with it next week!

src/Resolvers/DataFromSomethingResolver.php
rubenvanassche: Mhh, I get the point what you're doing, but running things through the pipeline is quite performance heavy and could have a lot of side effects since we're doing validation, mapping, authorization, casting and a lot of other things over there. 

I'm trying to find out how to solve this since we're now passing the data twice per morphable object through the pipeline.

We've got some options:

A) Don't run the pipeline at all or any other code, it is up to the user to decide what to do based upon the provided payload. This is the easiest solution from a package perspective, but the user should probably perform a lot of work payloads can be arrays, requests, ... property names may not be mapped, enums might not be casted, ...
B) Run the default pipeline but replace some pipes like Validation and Authorization which we don't need at this point. People could have replaced certain pipes with their own implementation, added pipes with side effects and we're still running a complete pipeline just to find out the class we actually need to send through the pipeline.
C) Run our own mini pipeline with the pipes like mapping names, casting, ... this basically has all the same problems as B.
D) Since we already tag properties as required for morph, why not add some hard constraints to them when provided to the morhph method like: payloads will be normalized in order to only work with arrays and thus provide an array as parameter, they won't be validated, property names will be mapped, it will only be allowed to use a string, int or enum as property type (we'll enforce that when creating the DataProperty) and enums will be cast if possible.
E) YOLO, we just run the pipeline


At the moment I'm leaning heavily towards D since we try to reduce the problem and side effects as much as possible. Some code will be duplicated but it wont be that much, I think I'm going to start working on an implementation of that part.

bentleyo: @rubenvanassche is there anything I can do to help out or do you have it sorted? Happy to put some extra time into it.

rubenvanassche:
@bentleyo 

Better late than never, somehow wasn't able to push to this PR, so created a new one with my changes: [CreationTest.php](https://github.com/spatie/laravel-data/pull/995)

If it looks good to you, we can merge this and release it! 

Tests fail due to some date formatting in non PR related tests, have to look at this later but non issue for now. Same for PHPStan.

bentleyo:
@rubenvanassche will close this one in favour of your new PR!

PR #928, opened 2025-01-11
Use BaseData contract i/o Data concrete in CollectionAnnotation

riesjart:
Use `BaseData` contract instead of `Data` concrete to determine if value type is data in `CollectionAnnotation`. This allows to read the annotation of collections that contain values of `Spatie\LaravelData\Dto` and other custom `BaseData` contract implementations.


rubenvanassche:
Thanks

PR #930, opened 2025-01-14
Add phpdoc to `TransformableData.php`

matthiasweiss:
I was running into an issue with phpstan on level 10, when I try to create an object using the `toArray()` method.

The code I have looks like this: 
```
Model::create($data->toArray())
```

This results in the following error:

```
Model>>::create() expects array<string, mixed>, array<mixed> given.
```

Adding phpdoc to the definition of the class should fix this issue. 
Since I was already touching this concern, I decided to add the phpdoc to the other methods returning arrays as well.


rubenvanassche:
Technically it kan return array<array-key, mixed>, can you change that?

matthiasweiss:
Alright, seems as if this is far more complex than I'd have hoped :(

PR #931, opened 2025-01-15
Allow disabling default Optional values in CreationContext

ragulka:
DTOs that have properties using the `Optional` type can be a little painful to work with when creating the DTO in a context where you'd rather have a missing value be resolved to `null`. This can happen when preparing a DTO on the server-side, later to be converted to a JSON payload. For instance, when an Eloquent model is converted to array, it may not include all the keys for the DTO - which would result in `Optional` type being used for the matching properties.

This PR makes it possible to control whether the `DefaultValuesPipeline` will set properties that support the `Optional` type to `Optional`, or fall back to `null` (if allowed by the property).

See the docs diff for usage.

### DIFF
--- a/docs/as-a-data-transfer-object/factories.md
+++ b/docs/as-a-data-transfer-object/factories.md
@@ -81,6 +81,29 @@ SongData::factory()
     ->from(['title' => 'Never gonna give you up', 'artist' => 'Rick Astley']); // album will `null` instead of `Optional`
 ```
 
+Note that when an Optional property has no default value, and is not nullable, and the payload does not contain a value for this property, the DTO will not have the property set - so accessing it can throw `Typed property must not be accessed before initialization` error. Therefore, it's advisable to either set a default value or make the property nullable, when using `withoutOptionalValues`.
+
+```php
+class SongData extends Data {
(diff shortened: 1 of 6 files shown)


rubenvanassche:
Can you add a tests which checks that indeed the optional value is replaced with null and also checks what happens when the option is on and we've got a type like string|Optional which doesn't support Optional?

ragulka:
> Can you add a tests which checks that indeed the optional value is replaced with null and also checks what happens when the option is on and we've got a type like string|Optional which doesn't support Optional?

@rubenvanassche sure thing - added in [b22cc7b](https://github.com/spatie/laravel-data/pull/931/commits/b22cc7b3fc0216c5f1322cfa58e65ff4c5218c02).

Note that when using `string|Optional`, the property will end up not being set. When converting such DTO to array, it will simply exclude that property from the array - which I believe is to be expected.

However, the property must not be accessed before initialization, or PHP will throw an error:
```
Error: Typed property SongData::$slug must not be accessed before initialization
```
This might throw off some, so I've added a warning to docs about this as well..

rubenvanassche:
Great PR, thanks!

PR #939, opened 2025-01-24
Add support for inertia deferred props

matthiasweiss:
Hello 👋

I've drafted up an initial version of support for inertia v2 [deferred props](https://inertiajs.com/deferred-props).

Usage looks like this:

```
class AutoDeferredData extends Data
{
    #[AutoInertiaDeferred]
    public InertiaDeferred $deferred;
}

$data = AutoDeferredData::from(['string' => Inertia::defer(fn () => 'Deferred Value')]
```

I'm not really happy with the implementation, since the concept of deferred props is technically not related to lazy properties, but the implementation is basically copied 1:1. On the other hand, this kind of makes sense, since the effects of lazy and deferred properties are very similar. 

My other approach would've been to basically copy the structure of the `abstract Lazy`, `AutoLazy` & `InertiaLazy` classes, but this would've required a bunch more changes.

However, I still don't really like how I've done it here, but I don't think this can be done a lot better without larger refactoring. Maybe a bigger refactoring even makes sense, since [merging props](https://inertiajs.com/merging-props) should also be implemented.

 I'm definitely open for any suggestions!

Best,
Matthias


src/Support/Lazy/InertiaDeferred.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Spatie\LaravelData\Support\Lazy;
+
+use Inertia\DeferProp;
+
+class InertiaDeferred extends ConditionalLazy
+{
+    public function __construct(
+        mixed $value,
```
matthiasweiss, line 10: This is `mixed` instead of `DeferProps', since the `$value` parameter in the `AutoInertiaDeferred::build` method is mixed and needs to be passed to this function

matthiasweiss:
@rubenvanassche I'm also not sure if this should really be part of the package.

The problem I had was that I wanted to achieve the following:

1) Have a data class that has one or more properties with the type `DeferProp` 

```
<?php

namespace App\Data;

use Inertia\DeferProp;
use Spatie\LaravelData\Data;
use Spatie\TypeScriptTransformer\Attributes\TypeScript;

#[TypeScript()]
class DashboardData extends Data
{
    public function __construct(
        /** @var array<PostData> */
        public array $myLatestPosts,
        public DeferProp|FeedData $feed
    ) {}
}

```

2) Convert all union types of type `DeferProp | SomethingData $property` to one of the following TypeScript types:

* `property?: SomethingData`
* `property: undefined | SomethingData`
* `property: null | SomethingData`

3) Use this data class as the second parameter to my Inertia response

```
<?php

namespace App\Http\Controllers;

use App\Data\DashboardData;
use App\Data\PostData;
use App\Queries\FeedQuery;
use Inertia\Inertia;
use Inertia\Response;

class DashboardController extends Controller
{
    public function __invoke(FeedQuery $query): Response
    {
        /** @var User $user */
        $user = auth()->user();
        $myLatestPosts = $user->posts()
            ->with('creator')
            ->latest()
            ->get();

        $data = DashboardData::from([
            'myLatestPosts' => PostData::collect($myLatestPosts)->all(),
            'feed' => Inertia::defer(fn () => $query->get()),
        ]);

        return Inertia::render('Dashboard/Dashboard', $data);
    }
}
```

By doing this, I can use the generated type as the type for the react component. This allows me to catch possible type mismatches between the data returned from the controller and the props expected by the component. If I change the name of the `myLatestPosts` variable to `yourLatestPosts`, the build will fail, since the react component still tries to access `myLatestPosts`, even though this property does not exist on `DashboardData` anymore.

```
export default function Dashboard({
    feed,
    myLatestPosts,
}: App.Dto.DashboardDto) { ... }
```

However, I just achieved this, more specifically generating a TS type where `property: null | SomethingData`, by simply adding the following mapping in the `typescript-transformer.php` configuration:

```
    'default_type_replacements' => [
        ...
        Inertia\DeferProp::class => 'null',
    ],
```

Not sure if there are any other use cases for supporting deferred props in this package.
Let me know what you think.

Best, 
Matthias

rubenvanassche:
Hi @matthiasweiss,

I'm not sure why you've closed this PR, implementation looks fine to me. It think we can even implement the merge functionality from Inertia like this.

As for TypeScript, if I'm not mistaken, if a property is Lazy then it will be an optional type in typescript so creating:

```php
class DashboardData extends Data
{
    public function __construct(
        /** @var array<PostData> */
        public array $myLatestPosts,
        public Lazy|FeedData $feed
    ) {}
}
```

Should be typed as

```ts
feed?: FeedData
```

We could even make it a bit more narrow by using:

```
class DashboardData extends Data
{
    public function __construct(
        /** @var array<PostData> */
        public array $myLatestPosts,
        public InertiaDeffered|FeedData $feed
    ) {}
}
```

Initialising the object would then happen using:

```php
DashboardData::from([
    'myLatestPosts' => PostData::collect($myLatestPosts)->all(),
    'feed' => Lazy::inertiaDeferredfn () => $query->get()),
]);
```

Let me know if there's something wrong with my understanding of the code 😄

matthiasweiss:
I just thought that it does not really make sense to add all of this code, if my use case can be covered otherwise!

But glad that you're still interested. Will look into it either tomorrow or the day after!

matthiasweiss:
@rubenvanassche has phpstan been added in the meantime or why does this suddenly fail?

rubenvanassche:
Updates of PHPStan often find new errors where it didn't find them earlier, I'll take a look at this later on, you don't have to fix them, we probably first need to fix main.

rubenvanassche:
Thanks!

PR #944, opened 2025-02-09
Introduce FromHeaderParameter Attribute: Map Header to DTO Prop

abdallah-zaghloul:
Map header input to data prop for some scenarios like: 
- send Binary Files using `HEAD` HTTP Method so other info should be sent via Headers
- map user insenstive headers like Language to DTO prop

### DIFF
--- a/src/Attributes/FromHeaderParameter.php
+++ b/src/Attributes/FromHeaderParameter.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Spatie\LaravelData\Attributes;
+
+use Attribute;
+use Spatie\LaravelData\Attributes\InjectsPropertyValue;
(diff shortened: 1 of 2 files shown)


rubenvanassche:
Could you

- write some tests like with the other attributes
- add some documentation so that other people will be able to use this feature
- remove the deprecated properties since these are new classes they can be omitted

abdallah-zaghloul:
> Could you
> 
> * write some tests like with the other attributes
> * add some documentation so that other people will be able to use this feature
> * remove the deprecated properties since these are new classes they can be omitted

Sure

rubenvanassche:
Since no changes were made I'm gonna close this one, feel free to to create a new PR with the changes.

PR #952, opened 2025-02-17
Fix data stub spacing

patrick-levesque:
This PR updates the default spacing in data.stub from 2 to 4 spaces to match the spacing used in other stub files. This is used when generating new data objects with the artisan command make:data.


freekmurze:
Thanks!

PR #953, opened 2025-02-18
Fix a display error in the docs

danielsimkus:
* Removed the > character as it doesn't seem to be used anywhere else either

Noticed an gapping issue in the docs, it displayed fine in PHPStorms markdown simulator, so maybe an underlying issue with the docs engine.

I think just removing the > character should fix it.

Screenshot 2025-02-18 at 19 47 14


freekmurze:
Thanks!

PR #955, opened 2025-02-20
Add EmptyString support for Formik compatibility

igorleszczynski:
# Add EmptyString support for Formik compatibility

## Description

Added support for generating Data Transfer Objects (DTOs) with empty string values instead of null values. This enhancement is particularly useful when working with Formik and Inertia.js in Laravel applications.

## Features
- New `EmptyStringData` trait and interface
- Extended `EmptyDataResolver` with empty string support
- Static method `emptyString()` that returns DTO with empty strings as default values

## Use Case
When working with Formik forms in React + Inertia.js + Laravel applications, you often need to provide initial values for form fields. Using null values with Formik can cause errors as Formik expects string values for its form fields. Previously, you would need to:

```php
// Before
UserStoreData::empty([
    'name' => '',
    'email' => '',
    'address' => '',
    // ... many more fields
]);
```
Now you can simply use:
```php
// After
UserStoreData::emptyString();
```
This is particularly beneficial for large DTOs where manually defining empty values for each field would be tedious and error-prone.
Example with [formik](https://formik.org/), [spatie/laravel-typescript-transformer](https://github.com/spatie/laravel-typescript-transformer) and inertia.js:

Controller:
```php
public function create(): Response
{
    return Inertia::render('Users/UserStorePage', [
        'user' => UserStoreData::form(),
    ]);
}
```

```tsx
const UserStorePage: React.FC<{user: UserStoreData}> = ({ user }) => {
    const handleUserStore = (values: UserStoreData, actions: FormikHelpers<UserStoreData>) => {
        /* store method content */
    };

    return <Formik initialValues={user} onSubmit={handleUserStore}>

    </Formik>
};

export default UserStorePage;
```
## Changes 
- Added `EmptyStringData` trait and interface
- Extended `EmptyDataResolver` with `defaultValue` parameter
- Updated base `Data` class to implement `EmptyStringData`

## Breaking Changes
None. This is a backwards-compatible enhancement.

### DIFF
--- a/src/Concerns/EmptyStringData.php
+++ b/src/Concerns/EmptyStringData.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Spatie\LaravelData\Concerns;
+
+use Spatie\LaravelData\Resolvers\EmptyDataResolver;
+
+trait EmptyStringData
+{
+    use EmptyData;
+    public static function emptyString(array $extra = []): array
+    {
+        return app(EmptyDataResolver::class)->execute(static::class, $extra, '');
+    }
+}
--- a/src/Contracts/EmptyStringData.php
+++ b/src/Contracts/EmptyStringData.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Spatie\LaravelData\Contracts;
+
+interface EmptyStringData
+{
+    public static function empty(array $extra = []): array;
+}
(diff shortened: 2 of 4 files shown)


rrmesquita:
This is a really valid use case. I don't use Formik no more, but React complains as well, when input values transition from undefined to something else, stating that the input went from uncontrolled to controlled.

rubenvanassche:
Could you rewrite this so that we don't add extra interfaces or traits? Just change the signature of empty to:

```php
public static function empty(array $extra = [], mixed $replaceNullValuesWith = null): array
```

Adding more methods to already packed data objects is something I want to skip.

igorleszczynski:
@rubenvanassche 
Now it should be ok

rubenvanassche:
Thanks!

Issue #962, opened 2025-03-01
php artisan data:cache-structures does not find my Data classes with reflection_discovery disabled - wrong conditions


cyppe:
**✏️ Describe the bug**

I need to include Data classes from a package. 

Following documentation https://spatie.be/docs/laravel-data/v4/advanced-usage/performance it should be possible to add my package to 'directories'. I disabled reflection_discovery.

``` 'structure_caching' => [
        'enabled'     => true,
        'directories' => [
            app_path('Data'),
            base_path('vendor/motoaction/motoaction-product-service/src/Data'),
        ],
        'cache' => [
            'store'    => env('CACHE_STORE', env('CACHE_DRIVER', 'file')),
            'prefix'   => 'laravel-data',
            'duration' => null,
        ],
        'reflection_discovery' => [
            'enabled'        => false,
            'base_path'      => base_path(),
            'root_namespace' => null,
        ],
    ],
````

But they are not found. 
I did some lightweight debugging and from what I can see it seems like either the condition is not correct, or it does not pick up the interface in the Data classes.

What I have tried so far:

vendor/spatie/php-structure-discoverer/src/Support/StructuresResolver.php

In execute method:

Adding dd() before it tests the conditions:

```

 dd( $profile->config->conditions, $structures);

        $structures = array_filter(
            $structures,
            fn (DiscoveredStructure $discovered) => $profile->config->conditions->satisfies($discovered)
        );

```

Then it shows that it actually does find my Data classes in both directories.  

But there is nothing in $structures that indicates the interface it tries to find in the conditions.

Dump:

```

Spatie\StructureDiscoverer\DiscoverConditions\ExactDiscoverCondition {#1086
  -conditions: array:1 [
    0 => Spatie\StructureDiscoverer\DiscoverConditions\ImplementsDiscoverCondition {#1098
      -interfaces: array:1 [
        0 => "Spatie\LaravelData\Contracts\BaseData"
      ]
    }
  ]
}


```
 

A few of the Data classes it finds (and later discards because it does not satisfies the conditions.

```

  "App\Data\TestData" => Spatie\StructureDiscoverer\Data\DiscoveredClass {#1304
    +name: "TestData"
    +file: "/Users/thomasandersson/PhpstormProjects/generate.motoaction/app/Data/TestData.php"
    +namespace: "App\Data"
    +isFinal: false
    +isAbstract: false
    +isReadonly: false
    +extends: "Spatie\LaravelData\Data"
    +implements: []
[...]

PR #963, opened 2025-03-01
Fix Data Class Discovery for Data Classes

cyppe:
This PR addresses the issue described in [#962](https://github.com/spatie/laravel-data/issues/962) where Data classes from external packages were not being discovered when added to the caching directories.

## Summary of Changes

- **Discovery Logic Update:**  
  Changed the discovery logic in `DataClassFinder.php` to check for classes extending the `Data` class instead of verifying an implementation of the `BaseData` interface.  
  This update ensures that both local and package Data classes are correctly identified and cached when `reflection_discovery` is disabled.

Fixes [#962](https://github.com/spatie/laravel-data/issues/962).

### DIFF
--- a/src/Support/Caching/DataClassFinder.php
+++ b/src/Support/Caching/DataClassFinder.php
@@ -2,7 +2,7 @@
 
 namespace Spatie\LaravelData\Support\Caching;
 
-use Spatie\LaravelData\Contracts\BaseData;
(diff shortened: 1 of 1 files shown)


rubenvanassche:
Solved by extending the chain: https://github.com/spatie/laravel-data/commit/a7d117e1224258a05c2d8e101928b6740a680a69

PR #966, opened 2025-03-04
add support for variadic parameter

joke2k:
Dear Reviewers,

I found a particular use case where I had a Data class with many parameters and two child classes extending it by adding more detailed parameters. Using the variadic parameter would have made it much easier to pass parameters to the parent constructor without replicating them in the child constructors. I believe this functionality was available in older versions (when reflection read the class properties instead of the constructor), and I looked for related issues but found nothing.
Let me know what you think and whether you believe it could be useful to include in the package.

**Current status**
```
class SimpleData extends Data 
    {
        public function __construct(public string $string)
        {
            parent::__construct(...$variadic);
        }
    }
class ExtendedData extends SimpleData
    {
        public function __construct(public string $name, ...$variadic)
        {
            parent::__construct(...$variadic);
        }
    }

$obj = ExtendedData::from([
        'name' => 'Wrapper name',
        'string' => 'SimpleData string',
    ]);
```

Currently raises this error:
```
the constructor requires 2 parameters, 1 given. Parameters given: name. Parameters missing: variadic
```

**Applying this fix**
```
$obj->name === 'Wrappr name';
$obj->string === 'SimpleData string'
```
as expected.

**Why support variadic parameters?**
It allows for easier extension of Data objects and passing parameters to the parent without specifying them again in the child constructor. Additionally, it enables the use of a standard PHP feature, avoiding the need to throw an exception.

### DIFF
--- a/src/Support/Factories/DataMethodFactory.php
+++ b/src/Support/Factories/DataMethodFactory.php
@@ -44,20 +44,7 @@ class DataMethodFactory
         ReflectionClass $reflectionClass,
         Collection $properties
     ): DataMethod {
-        $parameters = collect($reflectionMethod->getParameters())
-            ->map(function (ReflectionParameter $parameter) use ($reflectionClass, $properties) {
-                if (! $parameter->isPromoted()) {
-                    return $this->parameterFactory->build($parameter, $reflectionClass);
-                }
-
-                if ($properties->has($parameter->name)) {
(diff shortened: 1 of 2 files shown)


rubenvanassche:
Mhh not really sure if we want this since we try to create laravel-data as a package with strict typing which we kinda lose here. So gonna close this one.

PR #979, opened 2025-03-17
fix: Model with method name matching attribute names are not retrieved

### LINKED ISSUE #978: Model properties are not found if there is a method of the same name as of 4.14.0
I'm not sure exactly what is causing this issue, but since updating the following fails:

```cpp
UserData::from($user);
```
the constructor requires 8 parameters, 6 given. Parameters given: email, att, address, zipcode, city, country. Parameters missing: id, name.

Yet this works:
```cpp
UserData::from($user->toArray());
```

I'm confused what makes id and name special and unsure how laravel-data normally resolves the properties on a model.

Tofandel:
Fixes #978


rubenvanassche:
I'll fix PHPstan on main, looks like some tests are failing

AJenbo:
Thank you for addressing this so fast :)

PR #990, opened 2025-04-04
Update ide.json

Yi-pixel:
This is a configuration file for the Laravel Idea Plugin (for PHPStorm). With the correct configuration, it can automatically suggest properties defined in the Data when using certain methods.

The previous ide.json used classFqn, which seems not to work well with the latest version of Laravel Idea in PHPStorm. Upon reviewing the Laravel Idea schema, I found that classParentFqn exists and appears to be effective in my case.

Note that, according to the Laravel Idea plugin's [documentation](https://laravel-idea.com/docs/ide_json/completion#methods), classParentFqn may add some overhead.

Previously, it seems that only from/optional worked correctly, but now I've added several new methods, such as empty and some Permanently methods.

---

When using WithDeprecatedCollectionMethod, some new properties are added due to annotations. Using own here can be a good solution, but this will break #307, so I still plan to use all.

all:
image

own:

image

---

Here are some practical examples of usage:

image

image

image

image

### DIFF
--- a/ide.json
+++ b/ide.json
@@ -16,18 +16,31 @@
                     "place": "arrayKey"
                 },
                 {
-                    "methodNames": ["collection"],
+                    "methodNames": ["empty"],
+                    "classFqn": ["\\Spatie\\LaravelData\\Concerns\\EmptyData"],
+                    "place": "arrayKey"
+                },
+                {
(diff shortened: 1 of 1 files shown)


rubenvanassche:
Thanks!

PR #998, opened 2025-04-10
Fix #997

### LINKED ISSUE #997: 4.15.0 introduced breaking change for creating Data class with `null` value
**✏️ Describe the bug**
Previous to 4.15.x, we were able to create a Data class with a `null` value, as long as the Data class doesn't have any required values, ie:

```php
    public function __construct(
        public ?string $webhook_url,
        public ?string $level,
        public mixed $message,
    ) {}
```

**↪️ To Reproduce**

```php

it('can create data class with null value', function () {
    class BaseData extends Data
    {
      public function __construct(
        public ?string $name,
      ) {}
    }

    BaseData::from(null); // throws exception in 4.15, worked in 4.14
});
```

**✅ Expected behavior**
For a minor version increase, I would except the same behaviour as 4.14, where the class is created without a value set for `$name`

**🖥️ Versions**

Laravel: 12
Laravel Data: 4.15
PHP: 8.4

bentleyo:
Handle null payload.

This is doing the equivalent of this previous line:
https://github.com/spatie/laravel-data/blob/edd61b4dca5acdcfd1e3b7f2c19b75e83730f87c/src/Resolvers/DataFromSomethingResolver.php#L44

### DIFF
--- a/src/Resolvers/DataFromSomethingResolver.php
+++ b/src/Resolvers/DataFromSomethingResolver.php
@@ -41,7 +41,7 @@ class DataFromSomethingResolver
         $normalizedPayloads = [];
 
         foreach ($payloads as $i => $payload) {
-            $normalizedPayloads[$i] = $pipeline->normalize($payload);
+            $normalizedPayloads[$i] = $pipeline->normalize($payload ?? []);
         }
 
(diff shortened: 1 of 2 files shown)


rubenvanassche:
Thanks!

PR #1007, opened 2025-04-22
Add missing payload renormalization for morphDataClass

MartinPetricko:
Hey, so I am using abstract data with PropertyMorphableData contract implemented, but when correct data class is resolved, it only contains properties that are defined in that abstract class. This PR fixes that issue by renormalizing payload for correct morphDataClass. It is also a shame that this feature is not documented at all.

### DIFF
--- a/src/Resolvers/DataFromSomethingResolver.php
+++ b/src/Resolvers/DataFromSomethingResolver.php
@@ -48,6 +48,9 @@ class DataFromSomethingResolver
 
         if ($morphDataClass = $this->dataMorphClassResolver->execute($dataClass, $normalizedPayloads)) {
             $pipeline = $this->dataConfig->getResolvedDataPipeline($morphDataClass);
+            foreach ($payloads as $i => $payload) {
+                $normalizedPayloads[$i] = $pipeline->normalize($payload ?? []);
+            }
 
             $creationContext->dataClass = $morphDataClass;
             $class = $morphDataClass;


bcorcoran:
It is documented at https://spatie.be/docs/laravel-data/v4/as-a-data-transfer-object/abstract-data but the behavior you described is not. I just ran into this myself which is how I found the issue.

bcorcoran:
@rubenvanassche Any chance you could review this?

bcorcoran:
@freekmurze Could we get someone to review this?

bentleyo:
@MartinPetricko that's interesting 🤔 I haven't seen the behaviour you mentioned and I think there is a test for this at the moment: https://github.com/spatie/laravel-data/blob/main/tests/CreationTest.php#L1564

In that example the data class is created from a payload with data for `variant`, `a` and `enum`. Only `variant` is defined in the abstract class, but the data is created with the appropriate values for `a` and `enum`.

Are you able to provide an example that replicates the behaviour you mentioned?

MartinPetricko:
Hey @bentleyo thanks for response, this is my setup:

Package version: 4.15.1
Laravel version: 11.44.7

DataClasses:

    class BlockData extends Data
    {
        public int $id;
    
        public Type $type;
    
        #[LoadRelation]
        public ?BlockSettingData $settings;
    }

-

    abstract class BlockSettingData extends Data implements PropertyMorphableData
    {
        public int $id;
    
        #[PropertyForMorph]
        public int $blockId;
    
        {
            // returns settings data class-string
            return Block::findOrFail($properties['blockId'])->type->getDataClass();
        }
    }

-

    class TextSettingData extends BlockSettingData
    {
        public string $content;
    }

Now when I try to create a data object (`BlockData::from($blockModel)`) in my controller, TextSettingData will not fill `$content` property. It will only have `id` and `blockId`.

bentleyo:
@MartinPetricko thanks for providing more detail. This is definitely a bit of a complex situation!

I was able to replicate your issue by creating an [empty laravel repo](https://github.com/bentleyo/laravel-data-pull-1007/) with a simplified version. E.g. see `tests/Feature/ExampleTest.php` in:
https://github.com/bentleyo/laravel-data-pull-1007/compare/0a55549519f2e035d031ed576b44a9399c707bc5..f254e7e7669b31b341fd0319d222c8defde7f9a2

I can confirm the test fails as your described and your code change fix this behaviour. Hopefully this helps convince Spatie to implement it!

rubenvanassche:
@MartinPetricko Thanks for the PR, while your solution feels solid and is really simple it runs normalizers twice which feels kinda weird and could become a performance bottleneck.

I've implemented a new PR, fixing the root cause of the cause: #1041.

@bentleyo Thanks for the test repo, that test really helped me forward

@bcorcoran Please don't do this, we're all doing this in our free time, begging for merges is never going to work and just adds a lot of useless stuff to the conversation.

bcorcoran:
> @bcorcoran Please don't do this, we're all doing this in our free time, begging for merges is never going to work and just adds a lot of useless stuff to the conversation.

I apologize. I was too excited for the fix and had been using it already as a temporary fix. I appreciate the work you do.

PR #1012, opened 2025-04-30
Ensure toJson and toResponse call jsonSerialize

clementbirkle:
## Description

I noticed that the `toJson()` and `toResponse()` methods do not use the `jsonSerialize()` method.

This can be problematic in certain use cases. For example, you might override `jsonSerialize()` to transform an empty array into an object to ensure consistent data structures on the frontend.

With this PR, the package will be [more aligned with Laravel's behavior](https://github.com/laravel/framework/blob/12.x/src/Illuminate/Database/Eloquent/Model.php#L1745), which respects the `jsonSerialize()` method in similar scenarios.

### Example

```php
$data = new class extends Data
{
    public function __construct(
        public array $customFields = [],
    ) {}

    public function jsonSerialize(): array
    {
        $array = parent::jsonSerialize();

        // Ensure customFields is always an object
        $array['customFields'] = (object) $array['customFields'];

        return $array;
    }
};

// Current behavior:
json_encode($data);           // {"customFields":{}}
$data->toJson();              // {"customFields":[]}
$data->toResponse(request()); // {"customFields":[]}

// With this PR:
json_encode($data);           // {"customFields":{}}
$data->toJson();              // {"customFields":{}}
$data->toResponse(request()); // {"customFields":{}}
```

As shown, `json_encode($data)` and `$data->toJson()` currently produce inconsistent results. This PR fixes that.

### Implementation Note

To make `toResponse()` respect `jsonSerialize()`, I introduced a new method: `jsonSerializeWithTransformationContext()`. This method may also be useful in other parts of the package, such as:

- [Livewire integration](https://github.com/spatie/laravel-data/blob/main/src/Support/Livewire/LivewireDataSynth.php#L52)
- Possibly other areas as well?

### DIFF
--- a/src/Concerns/ResponsableData.php
+++ b/src/Concerns/ResponsableData.php
@@ -58,7 +58,7 @@ trait ResponsableData
         }
 
         return new JsonResponse(
-            data: $this->transform($contextFactory),
+            data: $this->jsonSerializeWithTransformationContext($contextFactory),
             status: $this->calculateResponseStatus($request),
         );
     }
--- a/src/Concerns/TransformableData.php
+++ b/src/Concerns/TransformableData.php
@@ -14,6 +14,8 @@ use Spatie\LaravelData\Support\Transformation\TransformationContextFactory;
 
 trait TransformableData
 {
+    protected null|TransformationContextFactory|TransformationContext $_jsonSerializeContextOnce = null;
+
     public function transform(
         null|TransformationContextFactory|TransformationContext $transformationContext = null,
     ): array {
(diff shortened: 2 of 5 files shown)


rubenvanassche:
Thanks, I'm not going to merge this PR because:

I don't see any reason to overwrite these methods since they will make it difficult to debug problems because you're changing how the package works.

Adding extra fields can be done with the with or additional functionality.

While I'm okay with toJson calling jsonSerialize, I don't want to add a specific method to pass through a TransformationContext into the JsonSerialize method to make it work with toResponse. This makes things too complicated.

clementbirkle:
Thank you, @rubenvanassche, for taking the time to review this PR and explain why it was closed.

I’m still a bit unsure about how to consistently convert an empty array attribute into an object when using the `toResponse` or `toJson` methods, while keeping it as an array when calling `toArray`.

I’ll go ahead and override these methods in my own project to achieve the desired behavior.

PR #1017, opened 2025-05-09
Add support for default values on properties for morph

bentleyo:
At the moment it isn't possible to add defaults to properties that are used to determine the morph for an abstract class.

While this might not be best practice, this limitation isn't mentioned in the documentation, and it feels like it _should_ work.
It turns out it's pretty easy to add so I think it might be worth including support for!

### DIFF
--- a/src/Resolvers/DataMorphClassResolver.php
+++ b/src/Resolvers/DataMorphClassResolver.php
@@ -58,6 +58,15 @@ class DataMorphClassResolver
                 }
             }
 
+            if ($dataProperty->hasDefaultValue) {
+                $morphProperties[$dataProperty->name] = $this->normalizeValue(
+                    $dataProperty,
+                    $dataProperty->defaultValue,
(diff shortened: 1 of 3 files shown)


src/Resolvers/DataMorphClassResolver.php
```diff
@@ -58,6 +58,15 @@ public function execute(
                 }
             }
 
+            if ($dataProperty->hasDefaultValue) {
+                $morphProperties[$dataProperty->name] = $this->normalizeValue(
+                    $dataProperty,
+                    $dataProperty->defaultValue,
+                );
+
+                continue;
```
bentleyo, line 67: Instead of setting `$found` to `true` I figured it was best to just continue to the next iteration of the loop since the next bit of code just checks that. However I see arguments for the other approach too.

rubenvanassche:
Thanks!

PR #1033, opened 2025-06-11
Implement the compare method on the eloquent casts to improve the isDirty check 

### LINKED ISSUE #1032: Eloquent cast is always detected as dirty
**✏️ Describe the bug**
When we use a Data object in a eloquent model cast, that field will always be marked as dirty even if nothing has changed.

**Explanation**

When retrieving a JSON field from the database it will result in a string `{"x": "hello world"}` note the space after the colon also the order of the fields can be different. This is the behavior of MySQL. And I also saw the issue here for Postgres  https://github.com/spatie/laravel-data/issues/496 

While the behavior of `json_encode` PHP results in the string `{"x":"hello world"}`. 

In eloquent itself there is a separated check for `object` and `collection` casts
https://github.com/laravel/framework/blob/12.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L2244

But this doesn't exists for custom casts, I created a PR in laravel https://github.com/laravel/framework/pull/55945 to add the ability to do custom value checks to detect if an attribute is dirty or not.

I hope it will be merged or another option will be created, after that this issue could be resolved by adding something like the following to the `DataEloquentCast` to fix this issue.

```php
    /**
     * Compare the two values, and check if they are equal
     */
    public function compare($model, string $key, $value1, $value2): bool
    {
        // Compare the two values loosly
        return json_decode($value1) == json_decode($value2);
    }
```

**✅ Expected behavior**
I would expect the `$model->isDirty()` to return false if nothing has changed.

**🖥️ Versions**

Laravel: 12.17.0
Laravel Data: 4.15.1
PHP: 8.4.7
Database: MySQL (ver. 8.0.36)

SanderSander:
Resolves #1032

### DIFF
--- a/src/Support/EloquentCasts/DataCollectionEloquentCast.php
+++ b/src/Support/EloquentCasts/DataCollectionEloquentCast.php
@@ -107,7 +107,7 @@ class DataCollectionEloquentCast implements CastsAttributes
 
     public function compare($model, string $key, $firstValue, $secondValue): bool
     {
-        return $this->get($model, $key, $firstValue, [])->toJson() === $this->get($model, $key, $secondValue, [])->toJson();
+        return $this->get($model, $key, $firstValue, [])->toArray() === $this->get($model, $key, $secondValue, [])->toArray();
     }
 
     protected function isAbstractClassCast(): bool
--- a/src/Support/EloquentCasts/DataCollectionEloquentCast.php
+++ b/src/Support/EloquentCasts/DataCollectionEloquentCast.php
@@ -105,6 +105,11 @@ class DataCollectionEloquentCast implements CastsAttributes
         return $dataCollection;
     }
 
+    public function compare($model, string $key, $firstValue, $secondValue): bool
+    {
+        return $this->get($model, $key, $firstValue, [])->toJson() === $this->get($model, $key, $secondValue, [])->toJson();
+    }
+
     protected function isAbstractClassCast(): bool
     {
         $dataClass = $this->dataConfig->getDataClass($this->dataClass);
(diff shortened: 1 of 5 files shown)


SanderSander:
> It looks like the Illuminate\Contracts\Database\Eloquent\ComparesCastableAttributes interface might be missing from your implementation.

This is on purpose, If the ComparesCastableAttributes class would be implement it would mean that the minimal laravel required would be version 12.18.0 or higher for this package.

By not implementing it, we can be sure that that everything is backward compatible, laravel would not check if the interface is implemented but does a simple check if the method exists on the custom cast class.

This would implicate that the dirty check would work when laravel version is 12.18.0 or higher, and if lower the package would work fine except the dirty check.

SanderSander:
@rubenvanassche  Done, I added a check on the laravel version so that the tests are skipped for older versions.

`version_compare($this->app->version(), '12.18.0') > 0 || $this->markTestSkipped('Laravel 12.18+ only');`

jhonaikerf:
Hi @rubenvanassche @SanderSander

From last version we have some erros with this implementation.

We are using nullable dtos, so the method toArray() dont apply to this.

Maybe add the nullsafe operator, which means if an attribute is nullable and its value changes from null to a non-null object.

@clementbirkle made the same comment

rubenvanassche:
Isn't this the case?

> For the nullsafe operator, I noticed that the method originalIsEquivalent in the laravel trait hasAttributes does check for null values, so for the implemention we can be sure that at least laravel doesn't pass null into the method

Otherwise feel free to PR it

PR #1035, opened 2025-06-12
Add the possibibilty to expcet key for empty data generation

thoresuenert:
In a scenario where you use sole properties to generate complex data you can come across a situation where you want to use `::empty()` to have a blueprint for your UI but cannot pass a value from the frontend.

For example:

```php
class TenantData extends Data
{
    public string|Optional $database;

    public function __construct(
        public string $name = 'Rick Astley',
    ) {
        $this->database =  Str::random(8);
    }
}

```

For the UI you want to use the `TenantData::empty()` to provide a blueprint.
With the current implementation this results in:
```php
$emptyData = [
    'database' => null,
    'name' => null,
];
```

With this PR you can pass an array with keys to except from the emptyData array like:
```php
$emptyData = TenantData::empty(except: ['database']);

$emptyData === [ 'name' => null ];
```

### DIFF
--- a/src/Concerns/EmptyData.php
+++ b/src/Concerns/EmptyData.php
@@ -6,8 +6,13 @@ use Spatie\LaravelData\Resolvers\EmptyDataResolver;
 
 trait EmptyData
 {
-    public static function empty(array $extra = [], mixed $replaceNullValuesWith = null): array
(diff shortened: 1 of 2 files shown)


rubenvanassche:
Thanks

PR #1046, opened 2025-06-24
Fix issue where toArray() is called on null

mdietger:
When a cast is null by default and we update it with a value an error is thrown.


rubenvanassche:
Thanks

PR #1048, opened 2025-06-24
can get the data class for a data collection by annotation with unicode characters

it4need:
When using data classes with unicode characters, the annotation resolving doesn't work properly, because the regex does not capture unicode chars.


Vmadmax:
Thanks!

PR #1049, opened 2025-06-25
Allow re-encrypting attributes when app uses previous keys

JaZo:
In #1033 the compare method is introduced, which is really nice! However, as Laravel now thinks the value is equal, it's impossible to "re-encrypt" the attributes without changing the actual value.

### Issue
Say you have a model that uses one of the casts provided by this package and have set it up to encrypt that data. Afterwards, you rotate the app key and set the app previous keys. Laravel handles decrypting the value using your new or old key, whichever works. Now you want to get rid of the previous key, maybe it's leaked, and have all data encrypted using your new key. With the "native" encrypted casts provided by Laravel, the value is re-encrypted using the new key and marked as dirty on update (see https://github.com/laravel/framework/blob/959fac8/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L2284-L2285 for details). So if you want to re-encrypt all data, you can just loop over all models, set the attribute to trigger a re-encrypt, and save them. However, since version #1033 introduced the compare function, Laravel thinks the values are equal and it doesn't mark the attribute as dirty and doesn't save the re-encrypted value.

### Solution
By implementing the same logic as Laravel, i.e. marking the attribute as dirty when there are previous encryption keys, we restore the ability to re-encrypt all data.

### DIFF
--- a/src/Support/EloquentCasts/DataCollectionEloquentCast.php
+++ b/src/Support/EloquentCasts/DataCollectionEloquentCast.php
@@ -107,6 +107,10 @@ class DataCollectionEloquentCast implements CastsAttributes
 
     public function compare($model, string $key, $firstValue, $secondValue): bool
     {
+        if (in_array('encrypted', $this->arguments) && ! empty(Crypt::getPreviousKeys())) {
+            return false;
+        }
+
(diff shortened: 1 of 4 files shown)


rubenvanassche:
That's a good catch! Thanks for the PR!

PR #1055, opened 2025-07-02
Add support for custom validation messages and attributes when using property morphable

bentleyo:
Custom validation messages and attribute names don't work properly when using property-morphable data at the moment. They build messages from the base class, not the morphed one, even when the morph attributes are provided.

This PR fixes that by making the `DataValidationMessagesAndAttributesResolver` class aware of property-morphable data.

There is a section of code I added that is now repeated in both:
- `DataValidationMessagesAndAttributesResolver`
- `DataValidationRulesResolver`

Not sure if it should be moved somewhere shared to avoid this duplication.

The code in question:

```
        if ($dataClass->isAbstract && $dataClass->propertyMorphable) {
            $payload = $path->isRoot()
                ? $fullPayload
                : Arr::get($fullPayload, $path->get(), []);

            $morphedClass = $this->dataMorphClassResolver->execute(
                $dataClass,
                [$payload],
            );

            $dataClass = $morphedClass
                ? $this->dataConfig->getDataClass($morphedClass)
                : $dataClass;
        }
```


ncphillips:
We're also running into this issue so I'm very excited for this fix. We make use of polymorphic collections in a couple places and ranging their error messages has been really frustrating.

ncphillips:
Actually, this doesn't seem to be a complete fix @bentleyo 

It works if I do `MyPolymorphicData::validate([ ... ])` but it does not work if I have a parent data object with a `Collection<MyPolymorphicData>` property on it.

Will investigate more later

ncphillips:
We've found a couple other problematic cases.

## Case 1: Nullable Polymorphic Child
The first is when an object has a **_nullable_** polymoprhic child and it happens to be null:
```ts
class Vehicle {}
class Car extends Vehicle {}
class Truck extends Vehicle {}

class Garage {
  vehicle?: Vehicle
}
```

**Note:** I opened a PR into this branch to try and fix it, but I've just realized it's failing. https://github.com/bentleyo/laravel-data/pull/3

## Case 2: Polymorphic Children of Non-Polymorpch Items in a Collection
The second case is more annoying. We changed our data structure to avoid the problem, but it's still worth mentioning.

If you have an array of non-polymorphic objects, but those objects themselves have a polymorphic child, then laravel-data will _not_ check any further down. For example, if I had a Region that contained various garages:

```ts
class Region {
  garages: Garage[]
}
```

This would fail to use the custom validation messages/attributes for the Vehicles in the Garage

rubenvanassche (changes requested):

rubenvanassche:
Thanks for the PR, a few changes are required. @bentleyo  Do we need changes to support the cases @ncphillips is mentioning? By the way thanks for the postcard!

bentleyo:
@rubenvanassche happy new year mate! When you get the chance could you please look over this one for me?

bentleyo:
@rubenvanassche is there anything needed on this PR before you'd be happy to consider this one?

I just ran into this problem again and had to add a temporary workaround in the parent of the property-morphable data class.

rubenvanassche:
Thanks!

PR #1069, opened 2025-08-07
Fix TransformableData docblocks

cosmastech:
I was getting static analysis errors here. There's nothing that guarantees the contents of all() has a string for a key.


rubenvanassche:
Thanks!

PR #1085, opened 2025-09-18
[feat] Add database constraints for validation annotations

Klaas058:
Currently when using annotations with database rules like `Exists` and `Unique`, it is not possible to include database constraints.

This will throw a php error:
```php
// Error: Constant expression contains invalid operations
#[Exists(Customer::class, column: 'id', where: fn($query) => $query->where('active', true))]
public ?int $customer_id
 ```
 
 This feature would allow using database constraints like so:
 ```php
 // Valid php 🎉
#[Exists(Customer::class, column: 'id', where: new WhereConstraint('active', true))]
public ?int $customer_id
```

After testing this has been working great for us, happy with any feedback

One thing I'm not a 100% sure about:
* Types in the `DatabaseConstraint` implementations are `mixed` since the Laravel methods are untyped themselves. Is this the best way?


docs/validation/using-validation-attributes.md
```diff
@@ -222,6 +222,41 @@ The rules will now look like this:
...
+        #[Unique('products', 'sku', where: [
+            new WhereConstraint('active', true),
+            new WhereInConstraint('type', ['physical', 'digital']),
+            new WhereNullConstraint('deleted_at'),
+        ])]
+        public string $sku,
+    ) {
+    }
+}
+```
```
rubenvanassche: Can you add a list here with the possible constraints?

src/Support/Validation/Constraints/WhereConstraint.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Spatie\LaravelData\Support\Validation\Constraints;
+
+class WhereConstraint implements DatabaseConstraint
+{
+    public function __construct(
+        public readonly mixed $column,
+        public readonly mixed $value = null,
```
rubenvanassche: I know indeed laravel has a mixed type but they provide typehints, let's use those. For value mixed is allright since it probably is really mixed. Another thing I would like to see is support for ExternalReference since we can dynamically replace the value which might be useful.

rubenvanassche (changes requested):
Looking great, few changes required and then this one is good to go!

src/Support/Validation/Constraints/WhereConstraint.php
Klaas058: Good suggestion! I've updated the PR to allow the use of `ExternalReference`. I also added types to the database constraints to the ones that I found made sense.

I took the type suggestions from https://api.laravel.com/docs/11.x/Illuminate/Validation/Rules/DatabaseRule.html 

I kept the `$value` of the `WhereConstraint` and `WhereNotConstraint` as `mixed`, like you said they probably really are mixed.

The type suggestion for `whereNot` does not include `null`, but that is supported.

This would otherwise error:
```php
new WhereNotConstraint('deleted_at', null)
```
Which is not a big deal, since you can use `WhereNullConstraint` instead, but it makes me think there are other inputs which are not in the type but are supported

rubenvanassche:
Thanks!

MrMeshok:
```php
#[Exists(Customer::class, column: 'id', where: static function ($query) {$query->where('active', true);})]
public ?int $customer_id
```
This actually works in PHP 8.5 thanks to https://wiki.php.net/rfc/closures_in_const_expr. But note what you need to use `static function` not `fn`

Issue #1100, opened 2025-10-18
[bug] Moph property of morphable class always has default value


litvintroll:
**✏️ Describe the bug**
Issue with `PropertyMorphableData` Data class
The nullable enum type field (default: null) is always `null` inside `morph()` method eventhough property was correctly resolved.
`DataMorphClassResolver` is not handling this case in a proper way.

**↪️ To Reproduce**

```php

it('can resolve morph class with backed enum type ignoring default value', function () {
    abstract class TestAbstractMorphableDataWithNullableBackedEnum extends Data implements PropertyMorphableData
    {
        #[PropertyForMorph]
        public ?DummyBackedEnum $type = null;

        public static function morph(array $properties): ?string
        {
            return $properties['type']?->value;
        }
    };

    $morph = app(DataMorphClassResolver::class)->execute(
        app(DataConfig::class)->getDataClass(TestAbstractMorphableDataWithNullableBackedEnum::class),
        [['type' => DummyBackedEnum::FOO]]
    );

    expect($morph)->toBe(DummyBackedEnum::FOO->value);
});
```
Debuging `\Spatie\LaravelData\Resolvers\DataMorphClassResolver::execute`:
Resolved property value
<img width="690" height="757" alt="Image" src="https://github.com/user-attachments/assets/fe5b5821-83b6-4a95-a0c6-a0d67daf16d6" />

replaced with default 'null'
<img width="690" height="748" alt="Image" src="https://github.com/user-attachments/assets/83da7a83-7d0f-44bc-a510-3c257a170b69" />

**✅ Expected behavior**
The default value must not be applied in case the field was correctly resolved.

**🖥️ Versions**

Laravel: 12
Laravel Data: 4.18
PHP: 8.4.4

rubenvanassche:
Thanks for the PR!

PR #1101, opened 2025-10-18
[fix] DataMorphClassResolver: skip default value applying if morph property was resolved

litvintroll:
issue: [Moph property of morphable class always has default value](https://github.com/spatie/laravel-data/issues/1100)


rubenvanassche:
Thanks!

PR #1108, opened 2025-10-29
added ability to more easily extend creation functionality

Ken-vdE:
Currently it's quite a hassle to overwrite or extend classes.
This change replaces `new` statements with `app()` to allow for easy extension and adds the ability to add `make` and `makeWithContext` static functions to your Data objects, so you can make use of the CreationContext for building the Data object or customise the creation further. 
That's great in combination with a possible extend on the CreationContext class, that's now possible.

### DIFF
--- a/tests/ValidationTest.php
+++ b/tests/ValidationTest.php
@@ -15,7 +15,6 @@ use Illuminate\Validation\Rules\In as LaravelIn;
 use Illuminate\Validation\ValidationException;
 use Illuminate\Validation\Validator;
 
-use Spatie\LaravelData\Support\Creation\CreationContext;
 use function Pest\Laravel\actingAs;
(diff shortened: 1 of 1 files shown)


rubenvanassche:
Thanks for the PR! For the creation side, the existing magic creation methods already cover this. If you define a `from*` method on your data class and type-hint `CreationContext`, it gets injected automatically. See the [factories docs](https://spatie.be/docs/laravel-data/v4/as-a-data-transfer-object/factories) for an example.

For customizing the validator, we don't support that through this pattern yet. We'd be open to a PR that adds a closure on `CreationContext` to configure the validator, that would be a more targeted solution than replacing internals with `app()` calls.

Is there a specific use case where the magic methods don't work for you?

PR #1118, opened 2025-11-27
BackedEnum to scalar cast

rajmundtoth0:
When serializing data often the target source expects `string|int` instead of a BackedEnum.


rubenvanassche:
Thanks for the PR! We try to keep data objects as strongly typed as possible, and a BackedEnum property is a good example of that. Converting it to a scalar would turn it into a union type which goes against that philosophy.

If you need the scalar value, you could add a transformer or use a PHP 8.4 property hook to derive it. I don't think we want to add extra casts for this use case.

PR #1124, opened 2025-12-09
Fix: use `findOrFail` instead of `firstOrFail`

christmex:
If we use `firstOrFail()`, it will throw this error:
```php
Illuminate\Database\QueryException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'some-ulid-or-id' in 'field list' (Connection: mysql, SQL: select `some-ulid-or-id` from `table_name` limit 1)
```


freekmurze:
Thanks!

PR #1159, opened 2026-03-08
Fix NormalizedModel not resolving non-snake-cased model attributes

### LINKED ISSUE #1148: [v4] Regression: Models that have non-snake cased columns are not normalized correctly.
**✏️ Describe the bug**
In version 4, when $snakeAttributes is true on the model (which is the default for laravel), creating a DTO from that model won't map non snake-case fields correctly, even if both the field on the model and the field on the DTO have the same casing. In version 3 of this package, it worked fine.

**↪️ To Reproduce**
```
<?php

uses(\Illuminate\Foundation\Testing\TestCase::class);

it('doesnt handle pascal case', function () {
    class MyData extends \Spatie\LaravelData\Data
    {
        public function __construct(
            public ?string $MyField,
        ) {}
    }

    class MyModel extends \Illuminate\Database\Eloquent\Model {}

    $myModel = new MyModel();
    $myModel->MyField = 'test';

    // In version 3:
    //   MyData^ {#3731
    //     #_partialTrees: null
    //     #_includes: []
    //     #_excludes: []
    //     #_only: []
    //     #_except: []
    //     #_additional: []
    //     #_wrap: null
    //     +MyField: "test"
    //   }
    //
    // In version 4:
    //   MyData^ {#4220
    //     #_additional: []
    //     #_dataContext: null
    //     +MyField: null
    //   }
    dd(MyData::from($myModel));
});
```

**✅ Expected behavior**
It should map the MyField property correctly like it did in version 3.

It actually looks like NormalizedModel shouldn't need to reference $snakeAttributes at all, because that property is ultimately only used for attributesToArray/toArray in laravel (and only for relations and mutated attributes at that), while NormalizedModel references the fields directly (whereas the version 3 one uses toArray() which is why it needs to check $snakeAttributes)
v4: https://github.com/spatie/laravel-data/blob/main/src/Normalizers/Normalized/NormalizedModel.php
v3: https://github.com/spatie/laravel-data/blob/v3/src/Normalizers/ModelNormalizer.php

**🖥️ Versions**

Laravel: 12.50.0
Laravel Data: 4.19.1
PHP: 8.2.26

isaackaara:
Fixes #1148

When `$snakeAttributes` is true (the default), `NormalizedModel::getProperty()` was unconditionally snake-casing the property name before looking it up via `getAttribute()`. This broke models that have PascalCase or camelCase column names — e.g. a model with `MyField` would get looked up as `my_field`, returning null.

This changes the lookup to try the original property name first, and only fall back to the snake_case version when the original name isn't found. This matches v3 behavior where `toArray()` handled attribute access directly.

### DIFF
--- a/src/Normalizers/Normalized/NormalizedModel.php
+++ b/src/Normalizers/Normalized/NormalizedModel.php
@@ -24,11 +24,25 @@ class NormalizedModel implements Normalized
 
     public function getProperty(string $name, DataProperty $dataProperty): mixed
     {
-        $propertyName = $this->model::$snakeAttributes ? Str::snake($name) : $name;
+        $propertyName = $name;
 
-        $value = array_key_exists($propertyName, $this->properties)
-            ? $this->properties[$propertyName]
-            : $this->fetchNewProperty($propertyName, $dataProperty);
+        if (array_key_exists($propertyName, $this->properties)) {
+            $value = $this->properties[$propertyName];
+        } else {
+            $value = $this->fetchNewProperty($propertyName, $dataProperty);
+
+            if ($value instanceof UnknownProperty && $this->model::$snakeAttributes) {
+                $snakeName = Str::snake($name);
+
+                if ($snakeName !== $name) {
+                    $value = $this->fetchNewProperty($snakeName, $dataProperty);
+
+                    if (! $value instanceof UnknownProperty) {
+                        $propertyName = $snakeName;
+                    }
+                }
+            }
+        }
 
         if ($value === null && ! $dataProperty->type->isNullable) {
             return UnknownProperty::create();


bastianschwarz:
Just as a heads up: I'm looking at our CI right now, seems like this also causes some cases where it did resolve snake case properties before to NOT be resolved anymore.

Looks like I have to add the snake case mapping explicitly now (which is good/correct) but might surprise some people and is kind of a breaking change.

Unless I missed something else.

wizzymore:
> Just as a heads up: I'm looking at our CI right now, seems like this also causes some cases where it did resolve snake case properties before to NOT be resolved anymore.
> 
> Looks like I have to add the snake case mapping explicitly now (which is good/correct) but might surprise some people and is kind of a breaking change.
> 
> Unless I missed something else.

Can confirm this adds breaking changes in my codebase too, some of the fields now don't get transformed from snake_case to camelCase.

rubenvanassche:
Hey all, we're going to revert this change. It breaks the automatic camelCase → snake_case mapping that most users rely on (as @bastianschwarz and @wizzymore reported).

Looking at this more closely, the original issue isn't really a bug. `$snakeAttributes = true` is a convention signal that tells Laravel your columns are snake_case. Laravel's own `toArray()` will also snake_case those keys. So having `$snakeAttributes = true` with PascalCase columns is a misconfiguration at the Laravel level, not something we should work around.

The claim that v3 handled this correctly is also a bit misleading. v3 used `toArray()` under the hood, which with `$snakeAttributes = true` would have snake_cased `MyField` to `my_field` too.

For models with non-snake-case columns, setting `$snakeAttributes = false` on the model is the right approach. For models with mixed casing, a `MapInputName` attribute is the way to go. We'll look into adding a model-specific mapping option down the road.

PR #1181, opened 2026-04-26
Support Inertia v3 by falling back to OptionalProp

jessegall:
Inertia v3 removed `LazyProp`. `Lazy::inertia` now returns `OptionalProp` when `LazyProp` is unavailable, preserving v2 behavior for consumer instanceof checks.

Bumps `inertiajs/inertia-laravel` to `^2.0|^3.0`.


rubenvanassche:
Thanks!

PR #1182, opened 2026-04-28
Add AutoWhenAppendedLazy attribute and related tests

netjordanlee:
Add attribute to support creating lazy properties for appended properties like accessors. Now accessor properties tagged with `AutoWhenAppendedLazy` attribute won't always be included when creating from a model unless explicitly appended.

### DIFF
--- a/src/Attributes/AutoWhenAppendedLazy.php
+++ b/src/Attributes/AutoWhenAppendedLazy.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Spatie\LaravelData\Attributes;
+
+use Attribute;
+use Closure;
+use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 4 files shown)


netjordanlee:
This was a feature I required in my own project, I had an accessor property that pulled in a computed form of a relationship, when passing in a model to the DTO when it checks for the property, because it is neither an typical attribute (column backed) or a relationship, it results in a fallback to getting the value - which means it would load the relationship, etc.

Because I don't always want this property to be loaded, only when I explicitly call `->appends()`, I needed some way to emulate the Lazy behavior of only including a relationship when loaded, but for when appended.

I don't know if this is specifically the best way to implement it, nor am I certain about the nomenclature of it being an `AutoLazy` attribute, but the implementation as-is works for my needs, hopefully it would be valuable for others.

Issue #1079, opened 2025-08-29
ValidateAndCreate skips validations of fields


AKurmaleev:
**✏️ Describe the bug**
Rules are not being applied when used via the rules method.

**↪️ To Reproduce**

1. Create Data class extending Spatie Data
2. Add rules method to the class to declare the rules, including all fields
3. Add constructor without the attributes for the parameters, with default values for the fields
4. Call DataClass::validateAndCreate($payload)

**✅ Expected behavior**

The payload is validated against all rules given in the rules method

**Actual behavior**

Only the fields that were given in the payload are being validated if the constructor parameter with the same name has the default value

**Reason**

DataValidationRulesResolver has this method when compiling the rules:
```
    protected function shouldSkipPropertyValidation(
        DataProperty $dataProperty,
        array $fullPayload,
        ValidationPath $propertyPath,
    ): bool {
        if ($dataProperty->validate === false) {
            return true;
        }

        if ($dataProperty->hasDefaultValue && Arr::has($fullPayload, $propertyPath->get()) === false) {
            return true;
        }

        return false;
    }
```

**Explanation**

This breaks possibility to use 'sometimes' and set the default values in the constructor, allowing flexibility, on top of that, it breaks any conditional rules of Laravel like requiredIf, prohibitedIf and etc. Of course having this in the code might be an intended behavior yet it conflicts with many basic Laravel validation features I cannot not call it a bug.

**Solution**

Possible to override (I know this well), but could be better having not broken basic Laravel rules out of the box, changing the method like that temporarily:
```
    protected function shouldSkipPropertyValidation(
        DataProperty $dataProperty,
        array $fullPayload,
        ValidationPath $propertyPath,
    ): bool {
        return $dataProperty->validate === false;
    }
```
Other way around would be completely removing the method and rework the function where it is called and other functions around it as more elegant and permanent solution.

**🖥️ Versions**

Laravel: v12.26.3
Laravel Data: 4.17.0
PHP: 8.4

AKurmaleev:
@rubenvanassche ok, did not check on the value you've set by default, because the value set is incorrect there to be passed in the first place, another point that was not done due to reproduction steps, here is the one that will fail now because the default value set will validate:
```
it('throws validation error when given an invalid value', function (){
    $dataClass = new class  extends Data
    {
    public function __construct(
        public string $property = 'yauttoittovuyu-2527@yopmail.com',
    ) {}

        public static function rules(?ValidationContext $context = null): array
        {
[...]

PR #1189, opened 2026-05-28
Fix #1188 by calculating promoted property default values from ancestors constructors

### LINKED ISSUE #1188: Change to validation in 4.22.0 causes unexpected validation errors
**✏️ Describe the bug**
After upgrading the package to the latest version I started getting validation errors I feel are unexpected.
I tried numerous versions of `laravel-data` and have isolated this to changes introduced in 4.22.0.

For example, take this class definition:
```php
abstract class BaseExampleData extends Data implements PropertyMorphableData
{
    public function __construct(
        #[PropertyForMorph]
        public string $type,
        public bool $example_bool = true,
    ) {}

    {
        return match ($properties['type']) {
            'failing' => FailingExampleData::class,
            'working' => WorkingFailingExampleData::class,
            default => null,
        };
    }
}

class FailingExampleData extends BaseExampleData
{
    public function __construct(
        bool $example_bool = true,
    ) {
        parent::__construct(
            type: 'failing',
            example_bool: $example_bool,
        );
    }
}
```

In versions of `laravel-data` < 4.22.0 you could run this code without any validation error:
```php
FailingExampleData::validate(['type' => 'failing']);
```

This feels like expected behaviour because the parent class defines a default for the `example_bool` property.

From `laravel-data` version 4.22.0 and onwards this will start throwing a validation exception saying that `example_bool` is required.

Interestingly, this alternative definition of the child class (with `public` in constructor - again even though it's defined on the parent) does work:
```php
class WorkingFailingExampleData extends BaseExampleData
{
    public function __construct(
        public bool $example_bool = true,
    ) {
        parent::__construct(
            type: 'working',
            example_bool: $example_bool,
        );
    }
}
```

I checked to see if `laravel-data` recognises the default values:
```php
$failingDataClass = app(DataConfig::class)->getDataClass(FailingExampleData::class);
dump($failingDataClass->properties['example_bool']->hasDefaultValue);

$workingDataClass = app(DataConfig::class)->getDataClass(WorkingFailingExampleData::class);
dump($workingDataClass->properties['example_bool']->hasDefaultValue);
```

This is actually the same result `false` then `true`, for `laravel-data` versions both before _and_ after that version.

I had a quick look and it seems like when `\Spatie\LaravelData\Support\DataClass` is built it only checks the classes constructor. I'm not sure if it makes sense to see to check the parent class as well, but this could get complex quickly. E.g. How many parents do you check? Do you need to consider the default value changing from subsequent `parent::__construct` calls?

So I'm not really sure what to do here. I can change my data class definitions, but it feels like something that _should_ work and will be very easy to miss/forget in the future when creating more data classes.

**✅ Expected behavior**
I should be able to define a public property with a default value in the constructor of an abstract class. Any classes that extend it should know there's a default value and avoid triggering `required` validation errors for it.

**🖥️ Versions**

Laravel: 12.48.1
Laravel Data: Tested with versions from 4.19.0 onward
PHP: 8.5.2

bentleyo:
Fix #1188 by capturing default values from promoted properties in ancestors constructors.

See that [Github issue](https://github.com/spatie/laravel-data/issues/1188) for more details of why this functionality is required.

While it is technically possible to override the value passed into the constructor e.g.

```php
public function __construct(
    string $example = 'a different value to the parent default',
) {
    parent::__construct(example: $example);
}
```

This would quickly become very complicated to support as the constructor can execute arbitrary code and the parameters could have different names/orders etc.

I think it's a reasonable trade-off that we only consider the value at the time of _promotion_. This still allows the child to override the parent class default by promoting again (by adding `public`) e.g.

```php
public function __construct(
    public string $example = 'a different value to the parent default',
) {
    parent::__construct(example: $example);
}
```

### DIFF
--- a/src/Support/Factories/DataClassFactory.php
+++ b/src/Support/Factories/DataClassFactory.php
@@ -5,7 +5,6 @@ namespace Spatie\LaravelData\Support\Factories;
 use Illuminate\Support\Collection;
 use ReflectionClass;
 use ReflectionMethod;
-use ReflectionParameter;
 use ReflectionProperty;
 use Spatie\LaravelData\Attributes\AutoLazy;
 use Spatie\LaravelData\Contracts\AppendableData;
--- a/src/Support/Factories/DataClassFactory.php
+++ b/src/Support/Factories/DataClassFactory.php
@@ -154,17 +154,23 @@ class DataClassFactory
             return $reflectionClass->getDefaultProperties();
         }
 
-        $values = collect($constructorReflectionMethod->getParameters())
-            ->filter(fn (ReflectionParameter $parameter) => $parameter->isPromoted() && $parameter->isDefaultValueAvailable())
-            ->mapWithKeys(fn (ReflectionParameter $parameter) => [
-                $parameter->name => $parameter->getDefaultValue(),
-            ])
-            ->toArray();
+        // Promoted properties store their default on the constructor parameter, not the
+        // property declaration, so they are invisible to getDefaultProperties(). Walk up from
+        // the concrete class collecting them; first definition wins so closer ancestors take priority.
+        $promotedDefaults = [];
+        $class = $reflectionClass;
+        $constructor = $constructorReflectionMethod;
+        do {
+            foreach ($constructor?->getParameters() ?? [] as $parameter) {
+                if ($parameter->isPromoted() && $parameter->isDefaultValueAvailable() && ! array_key_exists($parameter->name, $promotedDefaults)) {
+                    $promotedDefaults[$parameter->name] = $parameter->getDefaultValue();
+                }
+            }
+            $class = $class->getParentClass();
+            $constructor = $class ? $class->getConstructor() : null;
+        } while ($class);
 
-        return array_merge(
-            $reflectionClass->getDefaultProperties(),
-            $values
-        );
+        return array_merge($reflectionClass->getDefaultProperties(), $promotedDefaults);
     }
 
     /**
--- a/tests/Support/Factories/DataClassFactoryTest.php
+++ b/tests/Support/Factories/DataClassFactoryTest.php
@@ -8,7 +8,8 @@ use Spatie\LaravelData\Support\DataProperty;
 use Spatie\LaravelData\Support\Factories\DataClassFactory;
 
 it('can create a data class with defaults from promoted properties', function () {
-    class PromotedPropertyData extends Data {
+    class PromotedPropertyData extends Data
+    {
         public function __construct(
             public string $promotedProperty = 'hello',
             protected string $protectedPromotedProperty = 'hello',
@@ -30,7 +31,8 @@ it('can create a data class with defaults from promoted properties', function ()
 });
 
 it('can create a data class with defaults from inherited promoted properties', function () {
-    abstract class ParentNestedPromotedPropertyData extends Data {
+    abstract class ParentNestedPromotedPropertyData extends Data
+    {
         public function __construct(
             public string $promotedProperty = 'hello',
             public string $promotedPropertyWithOverride = 'hello',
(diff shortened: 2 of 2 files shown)


Bosun18:
Your fix lines up with what I found digging into #1188, but the tests here don't cover the thing the issue actually reports. `DataClassFactoryTest` asserts `defaultValue` on the built data class, which is the mechanism rather than the symptom — nothing in this PR fails if `validate()` keeps throwing. I have four validation-level tests for that: the morph case from the issue, the promoted sibling that already worked, the same thing on a plain `Data` child, and a string variant. All four pass on `fix-1188`, and the full suite is green there too, 1326 passed. Can push them to the branch or send a PR against your fork, whichever suits.

Two things I ran into that probably belong in the description, since they widen what this PR does. It surfaced in 4.22.0 because of #1177: before that, bool properties never got `required`, since `BuiltInTypesRuleInferrer` called `$rules->removeType(RequiringRule::class)` in the bool branch and `RequiredRuleInferrer::shouldAddRule()` returned early on `BooleanType`. #1177 removed both. That reads correct on its own, a bool with no default should be required, but it also took the lid off this.

And bool isn't really part of the condition. The same pair of classes with `public string $some_string = 'default'` in the parent and `string $some_string = 'default'` (not promoted) in the child throws `The some string field is required.` on main, on the commit right before #1177, and on 4.19.0, which is the oldest tag I ran it against. No `PropertyMorphableData` needed either, plain `Data` is enough. So this doesn't only restore 4.21 behaviour for bools, it changes validation for plain string properties in the same shape too, and that has been the behaviour for a while. I'd expect int and float to go the same way, though string is the only one I actually ran.

One thing to untangle before this closes #1188: Patabugen's report in that thread is a different bug. Plain property, no inheritance, no promotion, so `hasDefaultValue` is already `true` there and this PR has nothing to fix. What breaks it is the `$overwrittenRules` check from #1187, which means it starts at 4.23.0, not 4.22.0 as the comment says — 4.22.0 still passes. I ran it on `fix-1188` and it throws `The object field is required.` just the same. That behaviour looks deliberate, #1187 closes #1079 which asked for the opposite, so I don't think anything should change here. But that report wants its own issue, not to be closed along with this one.

The case I keep going back and forth on is the one your own test pins down. In `NestedPromotedPropertyData` the child declares `string $promotedProperty = 'hello from child'` without promoting it, and the test asserts the default is `'hello'`. But constructing that class with no arguments gives you `'hello from child'`, because that's the child's parameter default and that's what gets passed up. So the data class's idea of the default and PHP's disagree for that property. I understand why you'd stop there, walking constructor bodies is not a road anyone wants to go down, but the child's signature reads as if it wins, and that seems like something to spell out.

Smaller thing, and I might be misreading it: `expect($dataClass->properties)->not()->toHaveProperty('protectedPromotedProperty')` doesn't guard anything. `properties` is a `Collection`, and `toHaveProperty` looks at object properties, not collection keys, so it passes for `promotedProperty` too, which is definitely in there. `->not->toHaveKey(...)` would be the one with teeth. What you're checking is true, protected promoted params really don't end up in `properties`, it's just the assertion that can't fail.

bentleyo:
Hi @Bosun18 nice, it looks like you've done a very thorough review of the problem. I haven't looked at this issue in months so to be honest I've forgotten a lot of the details and it seems like this package isn't getting much attention at the moment - I created the PR months ago and no review.

I'm happy for you to take over if you want to fork from my branch (or start your own) and I can close this PR.

Bosun18:
Thanks, that works for me. I'll fork from `fix-1188` and keep your two commits as they are — the fix is yours, I'm only rebasing onto main and adding the tests on top. I'll open it as a separate PR and link back here, so no rush closing this one until that's up.

On the child-signature thing I raised, I'm leaving your trade-off exactly as it is. Walking constructor bodies isn't worth it, and widening the scope now would only give this PR another reason to sit unreviewed.

bentleyo:
@Bosun18 has created a PR here https://github.com/spatie/laravel-data/pull/1202 which has this fix and some additional tests. Will close this PR for now 👍

PR #1191, opened 2026-06-02
Allow multiple conditional requiring rules

### LINKED ISSUE #647: Cannot use multiple requiring rules

ibobgunardi:
Fixes #647.

`PropertyRules::add()` currently deduplicates all `RequiringRule` implementations as a group. That means adding `RequiredWith` after `RequiredUnless`, or adding more than one `RequiredWith`, drops the earlier conditional required rules even though Laravel supports combining them.

This changes add/prepend deduplication so conditional requiring rules can accumulate, while generated/plain `Required` is still removed when an explicit conditional requiring rule is added. Explicit `removeType(RequiringRule::class)` still clears all requiring rules.

Tests:
- `php -l src\Support\Validation\PropertyRules.php`
- `php -l tests\Support\Validation\PropertyRulesTest.php`
- `php -l tests\ValidationTest.php`
- `php vendor\bin\pest tests\Support\Validation\PropertyRulesTest.php --no-coverage`
- `php vendor\bin\pest tests\ValidationTest.php --filter "multiple required rules" --no-coverage`
- `php vendor\bin\php-cs-fixer fix --dry-run --diff --allow-risky=yes --config=.php-cs-fixer.dist.php src\Support\Validation\PropertyRules.php tests\Support\Validation\PropertyRulesTest.php tests\ValidationTest.php`
- `php vendor\bin\phpstan analyse src\Support\Validation\PropertyRules.php --memory-limit=1G`

Local note: the full `tests\ValidationTest.php` file gets past the new regression, but one unrelated existing test fails on this Windows/PHP setup with phpDocumentor rejecting an anonymous-class FQSEN path.

### DIFF
--- a/src/Support/Validation/PropertyRules.php
+++ b/src/Support/Validation/PropertyRules.php
@@ -3,6 +3,7 @@
 namespace Spatie\LaravelData\Support\Validation;
 
 use Illuminate\Support\Arr;
+use Spatie\LaravelData\Attributes\Validation\Required;
 
 class PropertyRules
 {
@@ -21,7 +22,7 @@ class PropertyRules
 
     public function add(ValidationRule ...$rules): static
     {
-        $this->removeType(...$rules);
+        $this->removeDuplicateRules(...$rules);
 
         array_push($this->rules, ...$rules);
 
@@ -30,7 +31,7 @@ class PropertyRules
 
     public function prepend(ValidationRule ...$rules): static
     {
-        $this->removeType(...$rules);
+        $this->removeDuplicateRules(...$rules);
 
         $this->rules = Arr::prepend($this->rules, ...$rules);
 
@@ -48,6 +49,33 @@ class PropertyRules
                 }
 
                 if ($rule instanceof $class) {
+                    unset($this->rules[$i]);
+
+                    continue 2;
+                }
+            }
+        }
+
+        $this->rules = array_values($this->rules);
+
+        return $this;
+    }
+
+    protected function removeDuplicateRules(ValidationRule ...$rules): static
+    {
+        foreach ($this->rules as $i => $rule) {
+            foreach ($rules as $newRule) {
+                if ($newRule instanceof RequiringRule && ! $newRule instanceof Required) {
+                    if ($rule instanceof Required) {
+                        unset($this->rules[$i]);
+
+                        continue 2;
+                    }
+
+                    continue;
+                }
+
+                if ($rule instanceof $newRule) {
                     unset($this->rules[$i]);
 
                     continue 2;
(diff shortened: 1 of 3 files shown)


ahmed-rashad-alnaggar:
@rubenvanassche can you review this?

PR #1192, opened 2026-06-03
Ignore original names for validated mapped input

### LINKED ISSUE #1019: Validation Bypass via Original Property Name in Mapped Inputs
**✏️ Describe the bug**
If your application uses mapped property names and the property is not required — meaning it can be omitted(e.g., by being set to null, having a default value, lacking a 'required' validation rule, or being conditionally validated with rules like required_if, sometimes, etc.) — then the validator can be bypassed by using the original property name.

This issue arises because the validation logic only checks the mapped property name. When it determines the property is not required and no value is provided under the mapped name, it skips further validation rules.
However, during assignment, the property can still be set using either the mapped name or the original name. This inconsistency between validation (which only looks at the mapped name) and assignment (which accepts both) creates a loophole where validation is bypassed but the property still gets assigned — potentially with invalid or unsafe data.

**↪️ To Reproduce**

```php
class Page extends BaseData
{
    #[Min(1)]
    #[MapInputName('page.number')]
    public int $number = 1;

    #[Between(1, 200)]
    #[MapInputName('page.size')]
    public int $size = 10;

    public static function prepareForPipeline(array $properties): array
    {
        // unset($properties['size'], $properties['number']);
    
        return $properties;
    }
}

it('bypasses validator using original attribute name', function () {
    // no error
    $page = Page::validateAndCreate([
        'size' => 300,
    ]);

    // 300
    dd($page->size);
});
```

**✅ Expected behavior**
print 10(Do not use the original property name)
or 
Illuminate\Validation\ValidationException: The page.size field must be between 1 and 200.(Keep the original property name)

**🖥️ Versions**

Laravel: 11.44.1
Laravel Data: 4.13.2
PHP: 8.2.20

**Confusion Around Mapping Property Names**
If a property has a map setting, can its original property name still be used?
If so, it feels more like an alias rather than a strict mapping.

**Do not use the original property name** 
Image

However, this conflicts with the following test case:
Image

**Keep the original property name** 
Image

But this also conflicts with the following test case:
Image

**Related link**
https://spatie.be/docs/laravel-data/v4/as-a-data-transfer-object/mapping-property-names

ibobgunardi:
Fixes #1019.

This prevents mapped input properties from falling back to the original PHP property name when validated creation is being used. Without this, an input like `['size' => 300]` can bypass validation rules defined for the mapped input name `page.size` and still hydrate the property.

The change only removes the unmapped original property during validation-backed creation, so existing unvalidated `from()` behavior continues to accept the original property name.

Tests run:

- `php vendor/bin/php-cs-fixer fix --dry-run --diff --allow-risky=yes --config=.php-cs-fixer.dist.php src/DataPipes/MapPropertiesDataPipe.php tests/ValidationTest.php`
- `php vendor/bin/phpstan analyse src/DataPipes/MapPropertiesDataPipe.php --memory-limit=1G`
- `php vendor/bin/pest tests/ValidationTest.php --filter "original property names" --no-coverage`
- `php vendor/bin/pest tests/MappingTest.php --no-coverage`

Not run: full `tests/ValidationTest.php` on my Windows/PHP 8.3 setup, because it hits an unrelated phpDocumentor FQSEN parsing issue in `it can overwrite collection item rules with wildcard rules`.

### DIFF
--- a/src/DataPipes/MapPropertiesDataPipe.php
+++ b/src/DataPipes/MapPropertiesDataPipe.php
@@ -2,8 +2,10 @@
 
 namespace Spatie\LaravelData\DataPipes;
 
+use Illuminate\Http\Request;
 use Illuminate\Support\Arr;
 use Spatie\LaravelData\Support\Creation\CreationContext;
+use Spatie\LaravelData\Support\Creation\ValidationStrategy;
 use Spatie\LaravelData\Support\DataClass;
 use Spatie\LaravelData\Support\DataProperty;
 
@@ -25,6 +27,10 @@ class MapPropertiesDataPipe implements DataPipe
             }
 
             if (! Arr::has($properties, $dataProperty->inputMappedName)) {
+                if ($this->shouldRemoveUnmappedPropertyName($payload, $properties, $creationContext, $dataProperty)) {
+                    unset($properties[$dataProperty->name]);
+                }
+
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


ropi-bc:
any news on when we can get this in a published package? This prevents us from upgrading to laravel 13

PR #1199, opened 2026-07-06
Fix validation of keyed collections with dotted array keys

### LINKED ISSUE #1198: Dots inside array keys break validation of keyed collections
**✏️ Describe the bug**
When a `Data` object contains a keyed collection (`array<string, ChildData>` / `DataCollection`) and the array keys themselves contain literal dots, validation via `validateAndCreate()` (or request/`validate()`) fails. The array key is treated as a nested path separator, so the validator looks for the value at the wrong location and reports the fields as missing.

Plain creation with `from()` is not affected. It looks like defining an empty `rules()` function also resolves the issue. Looks like #1043 introduced the regression.

**↪️ To Reproduce**

```php
it('will validate a keyed collection whose keys contain dots', function () {
    $dataClass = new class () extends Data {
        /** @var array<string, SimpleData> */
        #[DataCollectionOf(SimpleData::class)]
        public array $items;
    };

    DataValidationAsserter::for($dataClass)
        ->assertOk([
            'items' => [
                'key.with.dots' => ['string' => 'first'],
                'another.key' => ['string' => 'second'],
            ],
        ])
        ->assertErrors([
            'items' => [
                'key.with.dots' => ['other_string' => 'wrong key'],
            ],
        ])
        ->assertRules([
            'items' => ['present', 'array'],
            'items.key\.with\.dots.string' => ['required', 'string'],
        ], payload: [
            'items' => [
                'key.with.dots' => [],
            ],
        ]);
});
```

**✅ Expected behavior**
Validating keys containing dots should be resolved correctly
**🖥️ Versions**

Laravel: 12
Laravel Data: 4.18.0+
PHP: 8.4

lazerg:
Fixes #1198

When a keyed collection (`array<string, ChildData>`) has array keys containing literal dots, validation reports the fields as missing. The static rules path added in #1043 builds the rule key from the raw array key, so `key.with.dots` gets treated as nested path segments (`items.key.with.dots.string`) instead of one key.

This escapes the dots in the collection key so the rule key becomes `items.key\.with\.dots.string`, matching how the dynamic `Rule::forEach` path already handles it.

### DIFF
--- a/src/Resolvers/DataValidationRulesResolver.php
+++ b/src/Resolvers/DataValidationRulesResolver.php
@@ -178,7 +178,7 @@ class DataValidationRulesResolver
         DataRules $dataRules,
     ): void {
         foreach ($collectionPayload as $key => $value) {
-            $itemPath = $propertyPath->property($key);
+            $itemPath = $propertyPath->property(str_replace('.', '\.', $key));
 
             if (! is_array($value)) {
                 $dataRules->add($itemPath, ['array']);
--- a/tests/ValidationTest.php
+++ b/tests/ValidationTest.php
@@ -856,6 +856,35 @@ it('will validate a nullable collection', function () {
         ]);
 });
 
+it('will validate a keyed collection whose keys contain dots', function () {
+    $dataClass = new class () extends Data {
+        /** @var array<string, SimpleData> */
+        #[DataCollectionOf(SimpleData::class)]
+        public array $collection;
+    };
+
+    DataValidationAsserter::for($dataClass)
+        ->assertOk([
+            'collection' => [
+                'key.with.dots' => ['string' => 'Never Gonna'],
+                'another.key' => ['string' => 'Give You Up'],
+            ],
+        ])
+        ->assertErrors([
+            'collection' => [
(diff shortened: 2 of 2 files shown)


ropi-bc:
Thanks for jumping on it

lazerg:
Hi @rubenvanassche, just a gentle follow-up. It has been two weeks since my last ping. Could you take a look when you have a moment? Thanks!

cativo23:
Found the same fix working #1198 and saw this after. Going to open a PR with the same one-line change plus a regression test, linked from here. All credit for the diagnosis is yours — if you'd rather reopen this one instead, happy to hand over the test.

lazerg:
Go ahead with #1206. No need to reopen mine, since yours already carries the regression test.

The diagnosis holds. #1043 builds the rule key from the raw array key, so a literal dot becomes a path separator. The `Rule::forEach` path escapes it already, which is why an empty `rules()` hides the bug. Escaping inside `resolveStaticCollectionRules` is the right spot, because `ValidationPath` is also used for MapName and rule keys that contain dots on purpose.

